PR #97, opened 2015-03-20
Added the missing NOT LIKE operator

gregaubert:
The NOT LIKE operator was missing. I added it to my fork, here is a pull request if you want to merge it to your repos.


agershun:
Aubert. hi! Thank you for modifications. 

I think, we can modify grammar file to replace line:
```
    NOT LIKE    return 'NOT_LIKE'
```
to 
```
    NOT\sLIKE return 'NOT_LIKE'
```
to allow more than one space between NOT and LIKE.

In this case these lines should be inserted before:
```
    \s         return;
```

I will merge your pull requrest tonight and will check. Unfortunately, I have no big experience in merging.

Andrey

gregaubert:
Ah yes, good idea, and the same should be done for NOT BETWEEN then.
You want me to update the pull request with it ?

agershun:
Exactly!

Could you write update and I will apply it?

Отправлено с iPhone

gregaubert:
Okay I'll do it now

agershun:
Merged and recompiled. Thank you!

PR #129, opened 2015-04-19
Rename bower.json to .bower.json

Jinkwon:
Bower need .bower.json, not bower.json
Please change this name.

Thanks.


agershun:
I did it manually in develop branch. Thank you!

agershun:
I just registered AlaSQL as a Bower package. May I ask you to verify, how it is installed?

Jinkwon:
@agershun 
It works fine. :) 
Thanks.



BTW, I think should have added some more ignore property to .bower.json.
Bower will download full codes, it is too heavy as well.

agershun:
Thank you! 
I have never used bower before. Can you explain:
1. Can I include in the package only one directory with 'dist'?
* /dist

2. Do I have to include node_modules as well? AlaSQL uses only XLSX package for Excel import-export procedures

3. Do I need to include .bower file?

PR #221, opened 2015-05-21
Include dependencies in package.json

### LINKED ISSUE #214: es6-promise
I have a big issue with last 0.6 version  Error : Cannot find module ‘es6-promise’

leebyron:
`require('es6-promise')` is called within the source, but es6-promise is not included as a dependency, so if you install alasql, you'll get a missing module error.

This fixes #214

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "example": "examples",
     "test": "test"
   },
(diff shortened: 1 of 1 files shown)


agershun:
Than you very much! Just fixed.

PR #260, opened 2015-05-29
Depend on xlsx NPM package instead of js-xlsx

losnikitos:
JS-xlsx NPM package is named ```xlsx```, not ```js-xlsx```.
(see https://github.com/SheetJS/js-xlsx#installation, https://www.npmjs.com/package/xlsx)

Having ```js-xlsx``` dependency causes NPM to fail while installing alasql:
2015-05-29 14-38-02 banners and alerts

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "jison":"0.4.15",
     "lodash":"3.7.0",
     "xlsjs":"0.7.15",
-    "js-xlsx":"0.8.0"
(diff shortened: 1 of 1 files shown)


agershun:
Guys, sorry for that... 

@mathiasrw I will merge all changes you will make into 0.1.9.

losnikitos:
Wow, you are fast.
Thank you, guys!

PR #378, opened 2015-07-30
Allow _ as a wildcard character in LIKE queries

workmad3:
The SQL standard uses `_` as a single-character wildcard. This PR adds in support for `_` as a single character by replacing it with `.` during Regexp construction.


mathiasrw:
You are awesome. I will look into why the test is failing so we can implement the changes.

workmad3:
I had a quick look at the travis failures. It seemed to be node build issues and dependency issues with node 0.4 to 0.8 rather than test failures. They may have been temporary failures but I couldn't see a way to trigger a rebuild without doing an extra push with an empty commit (which seemed... unappealing :smile: )

mathiasrw:
I have added a test file of the changes and it looks good. Please accept the pull request at https://github.com/generalassembly/alasql/pull/1

Would be awesome if you could make a pull request to develop...

workmad3:
Ah, you're using a develop branch! I'll close this PR and open a new one for develop.

Thanks for the test file too, all looking good :) Pulled in https://github.com/generalassembly/alasql/pull/1 now so shouldn't take long.

workmad3:
PR is now on the correct branch: https://github.com/agershun/alasql/pull/381

PR #381, opened 2015-07-31
Allow _ as a wildcard character in LIKE queries

workmad3:
(This is transferring https://github.com/agershun/alasql/pull/378 from a PR on master to a PR on develop. Apologies for not noticing this commit flow initially). 

The SQL standard uses _ as a single-character wildcard. This PR adds in support for _ as a single character by replacing it with . during Regexp construction.

Thanks to @mathiasrw for adding the extra test file for this!


mathiasrw:
:+1:

Issue #410, opened 2015-09-11
alasql undefined under ionic / iOS (Phonegap, cordova)


ThinklabsGbR:
alasql works fine in android, chrome, IE, FF, Safari, and even on Safari within iOS. But as an iOS app, alasql fails to load. Maybe a problem with webworkers? I also noticed the "alasql"-defintion around Line 90 in alsql.js gets called over a 100 times! Is this normal?

    // (...) Around line 89/90
    var alasql = function alasql(sql, params, cb, scope) {
        console.log("alasql defined!");

Simple check for iOS regarding the alasql not found error:

      var check = 0;
      var checkAlasql = function(id) {
                    check++;
                    console.log("checking alasql: Try "+check);
                    if(typeof alasql !== 'undefined'){
                        console.log("alasql was DEFINED");
                    } else {
                        console.log("alasql was UNDEFINED");
                        setTimeout(function(id){
                            checkAlasql(id);
                        },100);
                    }
                };
      checkAlasql(10);

(must be run under iOS as an ionic/Phonegap-app!)

mathiasrw:
Thank you for investigating. 

What version of the library are you using? Could I ask you to try the latest version from the develop branch ( https://github.com/agershun/alasql/tree/develop/dist ) and verify if the problem pesists? We have had some issues with the UMD wrapping of the code. 

Does iOS phonegap apps run javascript as a webworker? (sorry, I dont have much experience with phonegap)

pboos:
With the current develop version it is not working either. alasql is UNDEFINED.

If including alasql-worker.min.js then it seems alasql is defined. But it seems we are running into other problems there. And it should work as well without having to include alasql-worker.min.js.

Any idea what it could be?

pboos:
Found the problem!!!

Problem is this line:

```
if(typeof window.indexedDB.webkitGetDatabaseNames == 'function') {
```

it should be

```
if(window.indexedDB && typeof window.indexedDB.webkitGetDatabaseNames == 'function') {
```

Because on iOS it seems to crash because window.indexedDB = undefined

pboos:
Strange. There seems to be a check above that line that i missed.

```
if(typeof window !='undefined' && typeof window.indexedDB != 'undefined') {
```

But seems like this check is not correct?

pboos:
Continued research.

Seems like that line is not working correctly on iOS.

I checked the following in the iOS app:

```
> window.indexedDB
< null
> typeof window.indexedDB
< "object"
```

[...]

PR #413, opened 2015-09-14
Fix cordova on iOS

pboos:
Fix alasql for running in cordova on iOS. #410


mathiasrw:
We really need to get the change from pboos@656792d into the develop branch and not the master branch. I could just copy the single line changed into develop - but then you would not be credited with the edits.

Please provide a new pull request to the develop branch or let me know if I should just put it into the develop branch.

pboos:
done :)

PR #415, opened 2015-09-14
Update 15utility.js

bjornblomqvist:
Fixed bug where Date instance is turned into a string on update.

### DIFF
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -704,6 +704,10 @@ var cloneDeep = utils.cloneDeep = function cloneDeep(obj) {
         return obj;
     }
 
+    if(obj instanceof Date) {
+        return new Date(obj);
+    }
+
     var temp = obj.constructor(); // changed
 
     for(var key in obj) {


mathiasrw:
Thank you so much for taking the time to dig into the problem and comming up with a solution!

We really need to get this into the develop branch and not the master branch. 

I would put the change into develop but then you would not be credited - wich is not really fair for you. 

Please provide a pullrequest to develop or let me know if I should just put the change into develop.

bjornblomqvist:
I did plan to start by creating a test that triggers the bug but i have no idea how to get the build system and test running. It would be great if you added some documentation on how to build and run tests.

Best Regards Björn Blomqvist

mathiasrw:
You are awesome! 

I will get the following info into the readme for sure:

- To build: `npm run build` 
- Add test: get into the test folder - copy and rename test000.js and give it the next number of test files (fint the lastes testfile and +1)
- To run test: `npm test`

PR #424, opened 2015-09-22
Added dist to gitignore

bjornblomqvist:
Added dist to gitignore and fixed so that gulp creates files needed for tests on startup.


mathiasrw:
You are so awesome - thank you.

Issue #440, opened 2015-10-16
Webpack and alasql - documentation note?


nickdeis:
Hey alasql devs,
Alasql doesn't work well with webpack since alasql uses "exports" to detected whether or not it's in node.js or the browser (not ideal but understandable).
However, this causes issues with webpack, because the is analyzed in node.

I don't think this warrants a code change, but adding a documentation note to use [script-loader](https://github.com/webpack/script-loader) and load alasql as follows

```js
import alasql from "script!alasql";
```

Best,
Nick

mathiasrw:
Great - thanks for the pull request !

kiranps:
@nickdeis 

``` import alasql from "script!alasql"; ```

gives the following error

``` Uncaught TypeError: (0 , _scriptAlasql2.default) is not a function ```

``` import  "script!alasql"; `` works fine

**Comment by mathiasrw** [comment]
@kiranps

Nice !

I will update the documentation...

**Comment by mathiasrw** [comment]
@kiranps 

Do you use the `ignorePlugin` things mentioned in https://github.com/agershun/alasql#webpack or can that be avoided with the `import "script!alasql";` syntax?

**Comment by mathiasrw** [comment]
@kiranps 

Any news on this?

**Comment by kiranps** [comment]
@mathiasrw ```ignorePlugin``` fixed the issue, i also had to add ```net|tls```.

alasql is using some json file, webpack say i need a separate loader for that.

should i use a json loader or is there any way to remove those warning

**Comment by mathiasrw** [comment]
GLad to hear. 

> alasql is using some json file

are you asking it to store files? is it only from `require('alasql')`?

**Comment by kiranps** [comment]
i have not used any json-loader in my webpack config, that's why webpack is throwing this error.
using a json loader it will fix the issue, but i don't think there is a need for json files while working in browser

```
/~/mime-db/db.json
Module parse failed: /home/XXX/XXXX/node_modules/mime-db/db.json Unexpected token (2:40)
You may need an appropriate loader to handle this file type.
```

**Comment by mathiasrw** [comment]
Hmmm... I see.

Its a nice little 4th level nested dependency:

```
+-- alasql@0.2.7 
¦ +-- dom-storage@2.0.2 
¦ +-- es6-promise@3.2.1 
¦ +-- lodash@4.13.1 
¦ +-- request@2.73.0 
¦ ¦ +-- aws-sign2@0.6.0 
¦ ¦ +-- aws4@1.4.1 
[...]

PR #441, opened 2015-10-16
#440: Webpack and alasql - documentation note?

nickdeis:
For #440

Added note to use script-loader for alasql with webpack


mathiasrw:
:+1:

Issue #403, opened 2015-09-01
IE8 issue


umer-farooq:
I am using alasql.min.js on IE 8 browser,it throwing me an error 'Expected identifier', but it working fine on IE10, chrome and firefox, is it a compatibility issue or do you have the workaround for this

agershun:
Could you provide the sample of the code or SQL statement where AlaSQL raise the error?

umer-farooq:
Here is the sample below, you can include alassql.min.js and path on line 4

<html>
<head>
<title>Alas Hello World</title>
<script type="text/javascript" src="scripts/alasql.min.js"></script>
</head>
<body>
TEST JOIN
<span id="testjoin"></span>
<br/>
TEST MULTIPLE JOIN
<span id="testmultiplejoin"></span>
<br/>
TEST WHERE
<span id="testwhere"></span>
<br/>
TEST UNION
<span id="testunionjoin"></span>
<br/>
TEST LEFT JOIN
<span id="testleftjoin"></span>
<br/>
TEST AGGREGATE
<span id="testagg"></span>
<br/>
<script type="text/javascript">
        var cars = {'Models' : [{model:'Mazda', price:20000}, {model:'Toyota', price:25000}, {model:'Ford', price:22000},{model:'Suzuki',pirce:50000}],'Cars' : [{model:'Mazda', color:'Blue'}, {model:'Toyota', color:'Red'},{model:'Ford', color:'White'}]};

	   var employee = {'Employee' : [{emp_id:1,emp_name:'James',salary:2000},{emp_id:2,emp_name:'Jack',salary:4000},{emp_id:3,emp_name:'Henry',salary:6000},{emp_id:4,emp_name:'Tom',salary:8000}],'Department': [{dept_id:101,dept_name:'Sales'},{dept_id:102,dept_name:'Marketing'},{dept_id:103,dept_name:'Finance'}],'Register': [{emp_id:1,dept_id:101},{emp_id:2,dept_id:102},{emp_id:3,dept_id:103},{emp_id:4,dept_id:104}]};
	   
[...]

PR #444, opened 2015-10-18
alasql-403: Sympathy for IE8 developers

nickdeis:
Ticket: #403 

Fix was to add gulp-dereserve to the build.
Not sure how to test, but I used IE11'S IE8 VM and ran a few queries against it.


mathiasrw:
Nice spotted...

PR #472, opened 2015-11-29
Update bower.json

dvelasquez:
Added "main" attribute section in bower.json to make compatible with automatized include scripts (gulp, grunt).
This allow to plugins like https://www.npmjs.com/package/gulp-bower-files to cleanly include alasql as a dependency.


mathiasrw:
Great !

Thank you very much :+1:

Issue #527, opened 2016-01-11
Cannot export to XLSX when running worker


silverhawk184:
I found a new issue where alasql will not export while running in worker mode.

SkeLLLa:
As I can see this issue isn't solved yet. Am I correct?

mathiasrw:
You are correct.

PR #555, opened 2016-01-21
#527 Changed window to self

nickdeis:
#527 Changed window to self. Probably should do this elsewhere

### DIFF
--- a/src/833xlsx.js
+++ b/src/833xlsx.js
@@ -28,7 +28,7 @@ alasql.into.XLSX = function(filename, opts, data, columns, cb) {
 	if(typeof exports == 'object') {
 		var XLSX = require('xlsx');
 	} else {
-		var XLSX = window.XLSX;
+		var XLSX = self.XLSX;
 	};
 
 	/* If called without filename, use opts */


agershun:
I affarid that we need to do it more globally, then just in this function. We can start AlaSQL initiation with determination of mode/environment like: 
* Node / browser / worker / vm / others

And more, ```self``` occasionally has ```window``` value, I think it is a side effect, so we can have global variable, like: ```alasql.mode``` or ```alasql.global``` to keep globals.

nickdeis:
This will always return the global
```javascript
Function('return this')()
```
Doesn't work with CSP, which is gaining some traction, but you can reenable eval through directives.

nickdeis:
[getGlobal|https://jsfiddle.net/3wrman1p/] inspired by [System.global|https://github.com/tc39/proposal-global]

agershun:
:+1: I will analyze it tonight. Thank you for great idea!

nickdeis:
Cool, I'll close this for now.

nickdeis:
I would also argue that the problem of getting global and finding out which env alasql is in are dependent. Env have only one thing in common: certain native functions on the global scope.
[Here's an attempt at figuring out which env alasql is in|https://jsfiddle.net/3wrman1p/1/]

agershun:
Let's open new issue and solve this problem in complex.

PR #569, opened 2016-01-30
LocalStorage engine - fix reDELETE with same WHERE clause

martinstarman:
See test for more info. Current behavior is that last line `alasql("DELETE FROM ls429.one WHERE str = 'a'")` just delete row with `str='b'`. This PR fix that.


mathiasrw:
Looks really good - thanks for taking the time to look into what was wrong !

martinstarman:
Thanks for quick response!

PR #581, opened 2016-02-03
Time conversion fix for 8 and 108 formats

Wisehorn:
It seems like a copy-paste typo got into getMinutes from getMonth.


mathiasrw:
Thank you!

Issue #583, opened 2016-02-03
Alasql in browserify setup in the browser


nowzig:
When doing importing or exporting of files (e.g. Excel) there is always an error like Uncaught TypeError: fs.writeFileSync is not a function

I use alasql with browserify in the browser.

I think it can be fixed by changing the checks like
`if (typeof exports === 'object')`
to 
`if (typeof exports === 'object' && process.browser !== true) `

Maybe there is a better solution?

mathiasrw:
Good point. The problem is found [here](https://github.com/agershun/alasql/blob/797b84d31f5e2c153925855c1b163552be033aaf/src/15utility.js#L469) - and your suggestion to a change makes a lot of sense. 

In #562 we are looking into making a central place to check for enviroment, so when its working we will replace the `if (typeof exports === 'object')` with `if (isNode())` so that we also support browserify 

Would i be ok if I get back to you when we need to get it tested in browserify?

agershun:
yes, agree.

NB: It is better to check if the ```process``` object availabe before: ```(typeof process === "object" && process.browser !== true)```

nowzig:
Great, that you are already working on a solution based on a central environment detection.
I would be happy to test it.

Yeah, checking if the `process` object is available makes sense.

nickdeis:
NB: You can exclude modules in browserify config by marking them as "external resources" or by pointing to an empty js file and aliasing it to that file.
@mathiasrw + @agershun 

Using isNode won't fix this, as both Webpack and Browserify use the AST and just pick up whatever has "require".
For Webpack I finally have a better solution this script-loader (and have been meaning to make a gist of it), but the other choice is to provide a build with all the requires stripped out.

nickdeis:
Exclude works well enough.

mathiasrw:
@nowzig Could you please retry again with the last version of [alasql.js](https://raw.githubusercontent.com/agershun/alasql/ef0bf8289e38f6de52a4fae85888c213bf0dfd8e/dist/alasql.js)?

nowzig:
First of all thank you for the fast corrections.

But now I get the error `Uncaught Error: Please include a Promise library`. The error comes from  [here](https://github.com/agershun/alasql/blob/594527fa7d31b3582c729091d9360b7d96eb3ab6/src/18promise.js#L11) 

Could it be because the check is for `object` but `window.Promise` is of type `function`?

mathiasrw:
[...]

PR #584, opened 2016-02-04
583: Added guides for Webpack and Browserify

nickdeis:
And it seems like my editor removes trailing whitespace


nickdeis:
For ticket #583

mathiasrw:
Wow - I really really dont get whý only the changes in c680497 are merged, but not the 2687f3a edits. 

Ill copy them over manually...

PR #638, opened 2016-03-24
"preIndex" and "helpdocs" variables are now declared.

chaddjohnson:
Hello!

My application depends on alasql via Bower. I am using grunt-contrib-uglify to uglify my code, and it was failing in alasql due to `preStart` and `helpdocs` not being declared via `var`. I have fixed this, and I suggest merging in this pull request.


mathiasrw:
Nice - thanks :)

PR #656, opened 2016-04-17
Adding isIndexedDB util

nbdamian:
Changing IndexedDB Check to use new util
Throw error if IndexedDB does not exist.

There is an issue with it causing alasql to be undefined. Unsure what is causing it.


nbdamian:
Should we be returning the `throw new Error` somewhere else in the script?
Where it is in my pull request is breaking the script, although where else could it be placed?

PR #681, opened 2016-05-26
Use gulp-jison instead of shell and make jison compiling cross-platfo…

vellotis:
…rm compatible.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,6 +9,7 @@ var closure = false;
 
 var fs = require('fs');
 var gulp = require('gulp');
+var jison = require('gulp-jison');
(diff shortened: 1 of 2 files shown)


vellotis:
Now that is really strange. Previously it didn't fail for node 0.12 when `parser` was used as module name. As the module name was wrong it should have failed completely for all node versions.

Now when I fixed it test errors occurred only in node 0.12. @mathiasrw Could you rerun travis tests?

BR

mathiasrw:
Awesome - thanks...

Its all good now. We might have a timing issue for the file test/604.json letting the test fail sometimes... (so lovely...)

Issue #598, opened 2016-02-14
Support for : ALTER TABLE ADD FOREIGN KEY ?


wadi-chemkhi:
Hello and thank you for sharing this great project.

I find myself often in need of using ALTER TABLE ADD FOREIGN KEY, are there any plans to include support for it ?

Thank you.

mathiasrw:
Any inputs @agershun ?

agershun:
Yes, we can... the only problem is time.

PR #682, opened 2016-05-26
Implement ALTER TABLE ADD/DROP FOREIGN KEY constraint

vellotis:
Solve #598.
Additionally improve
- ALTER TABLE DROP
- AlterTable toString

:information_source: **This PR is not finished! Need guidance!**

I need a guidance from someone who knows the source. It is a awesome project. BUT the code itself is IMHO a bit misorganised.

What are the aspects/requirements I would have to consider to call this feature implemented?

eg. indexes, primary key, unique, etc Which cases I have to handle? What would be the concept for writing tests?

BR


mathiasrw:
**Nice work !**

> It is a awesome project. BUT the code itself is IMHO a bit misorganised.

100% agree :+1: 



> What are the aspects/requirements I would have to consider to call this feature implemented?

We need to cover all new features with tests - so we make sure someone (like me) does not break what was already working. I suggest we make a copy of the testfile for altering table name and column name [test/test126.js](https://github.com/agershun/alasql/blob/develop/test/test126.js) and rename it to the next number in the row of testfiles



> eg. indexes, primary key, unique, etc Which cases I have to handle?

Start with a test of the ones you already implemented, when that is working we can have a look at [a full implementation og `ALTER TABLE`](https://msdn.microsoft.com/en-us/library/ms190273.aspx) and talk about what could be useful to implement.

vellotis:
Thanks @mathiasrw

I was currently concerned about Postgres+MySQL in my implementation. Maybe compliency flags would come in handy here?
ALTER TABLE specifications:
- [Postgres](https://www.postgresql.org/docs/current/static/sql-altertable.html)
- [MySQL](http://dev.mysql.com/doc/refman/5.7/en/alter-table.html)/[MariaDB](https://mariadb.com/kb/en/mariadb/alter-table/)
- [MsSQL](https://msdn.microsoft.com/en-us/library/ms190273.aspx)
- [SQLite](https://www.sqlite.org/lang_altertable.html)
- [Oracle](https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_3001.htm)

If you take a look at the whole list it is way too much to implement them at once. Or is MsSQL the primary one?

mathiasrw:
The link was only to illustrate how many details ALTER TABLE contains - not to be the main target. 

The important part is to identify what areas are most useful to you and if there are any differences in the implementations in other DBs highlite them so we can take an informed descision on what direction to follow when implementing so the documentation can be helpful

mathiasrw:
Hi @vellotis 

Do you need help or inputs to make the testfiles for what you have already implemented?

mathiasrw:
Hi @vellotis - how can I help?

vellotis:
Hi @mathiasrw 

I am really keen to this project. But in 2 weeks time I think I don't have time. After that I would be glad to implement it.

I first wanted to use AlaSQL as a replacement for MySQL in testing. But now I see that it is not yet that full featured.

So I must appologise that my on duty project is more prioritized at the moment. But I will come back to it in 2 weeks like I said.

mathiasrw:
Its all good - I just want you to know that im here when you feel like digging deeper :)

mathiasrw:
Hi @vellotis 

How are you going? Would you like me to look into resolving the branch conflicts?

mathiasrw:
Hey @vellotis 

Do you need help to fix the merge errors?

vellotis:
Hi @mathiasrw 

My PR still lacks proper tests.

Actually I saw that it would need a total rewrite. It has too much of spaghetti magic that would need to be refactored.
Have you considered refactoring AlaSQL into multiple modules by using eg. ES6 and babel? No offence

mathiasrw:
Yes - for sure!

Best shot was from nickdeis with https://github.com/agershun/alasql/issues/445#issuecomment-159793432 where parently we got some structures going, but no one was really sure it was the right way to do it as we would love to have different builds with a core, and then other features, so you get the size of file based on what you need. 

If you have any experience in structuring a project like this I would love to hear how you think we should proceed. 

Would loooooove to get the structure more clean and neat...

mathiasrw:
Should I close this?

PR #692, opened 2016-06-03
Update type script definition

kaktus40:
I made some change in typescript definition file in order to use fn and aggr see [here](https://github.com/agershun/alasql/wiki/User%20Defined%20Functions)


gulpfile.js
```diff
@@ -231,7 +231,7 @@ gulp.task('copy-console-org', function(){
 
 // Echo plugin
 gulp.task('typescript', function(){
-  gulp.src(['./src/typescript/alasql.d.ts'])
+  gulp.src(['partners/typescript/alasql.d.ts'])
```
mathiasrw: 👍

mathiasrw:
Looks really good. Thank you so much for your help !

PR #712, opened 2016-07-20
Include check constraint names in violation errors

bopjesvla:
Instead of throwing a generic "Violation of CHECK constraint", a violation now throws "Violation of CHECK constraint {constraint_name}".


mathiasrw:
Thank you!

PR #713, opened 2016-07-20
Postgres-specific aliases and fixes

bopjesvla:
Adds ~ as an alias for REGEXP, \~~ and \~~* for LIKE, and !~~ and !~~* for NOT LIKE, see https://www.postgresql.org/docs/9.3/static/ddl-constraints.html.

Also, pg_dump adds some weird unnecessary conversions of obscure types, which is why I turned conversions to unknown types into noops.


mathiasrw:
This is great!

I can see you added 11 lines that is never reached by a test  scenario.

Can I ask you to make a new test file in `test/` asserting the new behavour? 

This way we can eliminate future problems by someone accidently breaking the functionality.  Have a look at https://github.com/agershun/alasql/blob/develop/test/test000.js to see an example.

bopjesvla:
Done!

bopjesvla:
I'm guessing that those 8 lines are lines I moved which didn't have tests before.

mathiasrw:
You are awesome - thank you so much...

PR #720, opened 2016-07-28
Allow unknown functions to be defined on alasql.fn afterwards

bopjesvla:
Right now, unknown SQL function calls are transpiled to empty strings, which results in cryptic errors. This PR solves this by transpiling unknown_function(a, b) to alasql.fn.unknown_function(a, b) regardless of whether alasql.fn.unknown_function is actually defined, while still respecting alasql's stdfns. This also allows folks to define the implementation of functions after the fact, which I require in one of my projects.

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -66,7 +66,23 @@ yy.FuncValue.prototype.toJS = function(context, tableid, defcols) {
 	var s = '';
     var funcid = this.funcid;
 	// IF this is standard compile functions
-	if(alasql.fn[funcid]) {
+	if(!alasql.fn[funcid] && alasql.stdlib[funcid.toUpperCase()]) {
+		if(this.args && this.args.length > 0) {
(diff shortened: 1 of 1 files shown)


mathiasrw:
Really good idea. 

Can I ask you to make a new test file in test/ asserting the new behavour?

This way we can eliminate future problems by someone accidently breaking the functionality. Have a look at https://github.com/agershun/alasql/blob/develop/test/test000.js to see an example.

bopjesvla:
It's a simplification of old behaviour so regression is basically impossible. As you can see, code coverage went up even though I didn't add any tests.

mathiasrw:
Super valid point. 

I always ask when no test file is changed / created for a PR

bopjesvla:
Can you do me a favour and publish this to npm as a patch? Then I won't have to remember to change my dependencies back.

mathiasrw:
Sure... 0.3.1 is up now.

PR #731, opened 2016-08-11
Add stub for ARRAY[]

bopjesvla:
pg_dump turns expressions like `any(1, 2)` into `any(array[1,2])`. This PR counteracts that, and could also function as a stub for full-fledged array support.


mathiasrw:
Thank you!

PR #732, opened 2016-08-12
Fix hasIndexedDB

ku:
It seems that `window.indexedDB` is `null` when it is not available. I've faced a problem alasql raises runtime error on ios8+chrome and found it is caused since `(typeof utils.global.indexedDB !== 'undefined');` returns true when `window.indexedDB` is `null`.


mathiasrw:
Awesome !

PR #734, opened 2016-08-15
Make string escape sequences SQL compliant

bopjesvla:
Right now, alasql is using JavaScript's escape sequences (`'\.'` is turned into `'.'`, for example). In SQL, the only characters that can be escaped are single quotes.

This PR is probably going to break stuff in some SQL dialects, but it's necessary for Postgres support.


mathiasrw:
Thank you so much for making the quality better.

It might break something - but as long as our regression tests are still OK its fine for now.

(and if any issue come up we fix it when time comes...)

PR #743, opened 2016-08-31
Implement three-valued logic

bopjesvla:
`(null = null) is null`, baby.

Concerns:

- I modified a regression test that asserted SQL-incompliant behavior so you may want to bump a minor.
- To check if any part of an expression is null, all subexpressions (usually the left and right side) are stored in a fresh array for every expression. This may slow things down, but I guess we'll see about that the next time someone runs benchmarks.


mathiasrw:
Thank you!

> I modified a regression test that asserted SQL-incompliant behavior so you may want to bump a minor.

As we are still in 0.x.y we try to keep changes in minor for big changes...

We will go to version 1.0.0 when we can get 95% success on [the SQLlogictest)(https://github.com/agershun/alasql/blob/develop/TESTLOG.md#sqllogic) that SQLite uses

PR #744, opened 2016-09-03
Better postgres arrays

bopjesvla:
Continuation of #731


bopjesvla:
hold up

bopjesvla:
that's all

mathiasrw:
You are a machine!

Thanks :)

mathiasrw:
Im trying to update the documentation a bit. 

I see  `ANY (array[1,2])` and `ARRAY[1,2,3]` in the new tests, and will include it in the description for array keyword, but I cant figure out what https://github.com/agershun/alasql/pull/744/commits/cd4f46bb56ca5f4e9257a8e02459d30139378769 and https://github.com/agershun/alasql/pull/744/commits/f10dd64df37f8105b38cdc9707d146e2c76587a7 is adding as functionality. 

Could you give me an example? I can add documentation and a test if you just give me a hint :)

bopjesvla:
The first two commits in this PR are kinda misguided and should probably be ignored. The commits you referenced are the real deal.

The first of those implements Postgres arrays. The difference between JSON arrays, which were already implemented, and pg arrays is that the former only takes JS primitives while the latter takes anything.

In a pg database, columns that take arrays have a type like `text[]`, which denotes an array of strings. The last commit simply makes sure that those column types are allowed.

bopjesvla:
Actually the first commit is also pretty sane

Again, I'd appreciate it if you could publish this to npm as a patch

mathiasrw:
> Again, I'd appreciate it if you could publish this to npm as a patch

Sure - consider it done...

PR #745, opened 2016-09-04
Allow non-reserved keywords as identifiers

bopjesvla:
Of SQL's trillion keywords, a large part can also be used as identifiers. Since Jison doesn't have a fallback mode for this type of thing, I hooked into the error event.


bopjesvla:
oh

PR #746, opened 2016-09-05
Add support for VALUE inside checks (a stub for domains)

bopjesvla:
Reference: https://www.postgresql.org/docs/9.5/static/sql-createdomain.html

> CHECK clauses specify integrity constraints or tests which values of the domain must satisfy. Each constraint must be an expression producing a Boolean result. It should use the key word VALUE to refer to the value being tested. Expressions evaluating to TRUE or UNKNOWN succeed. If the expression produces a FALSE result, an error is reported and the value is not allowed to be converted to the domain type.
> 
> Currently, CHECK expressions cannot contain subqueries nor refer to variables other than VALUE.

Because check constraints for domains only receive a single value, I made VALUE refer to the base context.


test/test231.js
```diff
@@ -62,7 +62,7 @@ describe('Test 231 NIST SQL Example', function() {
                     (ID INTEGER REFERENCES STATION(ID), \
                     MONTH INTEGER CHECK (MONTH BETWEEN 1 AND 12), \
                     TEMP_F REAL CHECK (TEMP_F BETWEEN -80 AND 150), \
-                    RAIN_I REAL CHECK (RAIN_I BETWEEN 0 AND 100), \
+                    RAIN_I REAL CHECK (VALUE->RAIN_I BETWEEN 0 AND 100), \
```
mathiasrw: This change can not be parsed at the moment: https://travis-ci.org/agershun/alasql/jobs/157699517#L816

Did you push the last version of the `dist/` folder?

bopjesvla: apparently not

mathiasrw:
Thank you for more awesome code

PR #752, opened 2016-09-14
export typescript definition for native import

jnath:

you can use import * as alasql from "alasql" 
juste install with npm.


mathiasrw:
Awesome!

Thanks :)

mathiasrw:
Ill add the change to the source :)

PR #764, opened 2016-10-06
Add missing var declaration

tkhyn:
Hello,

Just found another undeclared variable when bundling 0.3.2.
Will you want the changed dist files as well or do you compile them only at release time?

Thanks again for alasql !


mathiasrw:
Awesome - thanks :)

PR #777, opened 2016-11-06
Implement TRIGGER AFTER DELETE

AndreHeber:
I've implemented the feature to trigger a function after DELETE. There is no information about the change itself, only that it is changed. But that is exactly what I need in my own project.

### DIFF
--- a/src/60createtable.js
+++ b/src/60createtable.js
@@ -506,19 +506,6 @@ yy.CreateTable.prototype.execute = function (databaseid, params, cb) {
 				table.uniqs[uk.hh][ukaddr]=undefined;
 			});
 		}
-
-		// Trigger prevent functionality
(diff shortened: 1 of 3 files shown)


AndreHeber:
This is strange. `npm test` just runs fine!

kanghj:
@AndreHeber, I think you have to commit the contents of the `dist` directory too!

mathiasrw:
@AndreHeber Its because the test runs both from dist and from src (to make sure we dont push something that does not have build src)

I merge the change and build from src to make it all good...

AndreHeber:
Ah, nice.
It is my first pull-request, I'm happy now. 😀
And next time, I will commit dist too.
Thank you!

mathiasrw:
💃

Issue #741, opened 2016-08-26
Error when do JOIN between 2 CSV files...


tiagowippel:
Hi,

It´s possible to do a JOIN between 2 CSV ?

I have try to do this:

```
alasql("select * from csv('ITVPE.CSV',{headers:true}) itv \
 join csv('PROPE.CSV',{headers:true}) pro \
 on itv.codipro = pro.codipro",[],(data) => {
  console.log(data);
}); 
```

get this error on nodejs: **_Error: Unknown type of FROM clause_**

PROPE.CSV

```
CODIPRO,NOMEPRO
"01001","TESTE 01001"
"01002","ZTESTE 01002"
"01003","TESTE 01003"
```

ITVPE.CSV

```
NRNOTA,CODIPRO,QUANTI
123,"01001",10
123,"01002",20
123,"01003",30
122,"01001",30
122,"01003",50
```

mathiasrw:
There is no reason why this should not be possible.

There could be a bug in the parser.

What happens if you do a `... JOIN (SELECT * FROM csv('PROPE.CSV...`?

----

Bonus info: headers are set to true as default and params can be skipped, so your SQL could also be written as

```
alasql("select * from csv('ITVPE.CSV') itv \
 join csv('PROPE.CSV') pro \
 on itv.codipro = pro.codipro",(data) => {
  console.log(data);
});
```

mathiasrw:
@tiagowippel 

What happens if you do a ... JOIN (SELECT * FROM csv('PROPE.CSV'))...?

tiagowippel:
Hello again,

i have try this:


```
alasql(
    `select * from CSV('ITVPE.CSV',{headers:true,separator:','}) ITV
    join (select * from CSV('PROPE.CSV',{headers:true,separator:','}) PRO)
    on ITV.CODIPRO = PRO.CODIPRO`,
    (data) => {
        console.log(data);
    });
```



got this error:

**C:\Users\TI\dev\dev2\node_modules\alasql\dist\alasql.fs.js:2201
        throw new SyntaxError(str)
        ^

SyntaxError: Parse error on line 2:
...ator:','}) PRO)    on ITV.CODIPRO = PRO
----------------------^
Expecting 'LITERAL', 'BRALITERAL', 'AS', got 'ON'
    at Object.parser.parseError (C:\Users\TI\dev\dev2\node_modules\alasql\dist\alasql.fs.js:2201:8)
...**

mathiasrw:
[...]

PR #778, opened 2016-11-07
Change grammar for join

kanghj:
For #741. (I numbered the issue wrongly in the branch name, sorry if it causes confusion!)

The results of the query in the test (in `test434.js`, which I added) look wrong to me, but I haven't had time to look at it that carefully, and it looks like it might be a deeper problem.


mathiasrw:
This is awesome! thank you so much :)

Issue #793, opened 2016-12-06
Double quotation marks in CSV output #431 : getting same error on latest version ie 0.3.4


junedansari:
getting extra quotes on first column header while exporting CSV

like #431

mathiasrw:
Terrible - can you provide some code that generates the problem?

Dhananjay05414:
Hi Mathias,
Whenever using **"alasql": "0.2.6"** version it's working fine but I have updated **"alasql": "0.3.4"** then facing  double quotation issue in downloaded output CSV .Please find attached screenshot and code snippet.

  alasql.promise('SELECT * INTO CSV(\"'+fileName+'\",{headers:true,separator:","}) FROM ?',[dataToWrite])
![input_format_of_csv](https://cloud.githubusercontent.com/assets/24468450/21095601/3812bcfa-c082-11e6-9926-4442aeec35e5.png)
![output_format_of_csv](https://cloud.githubusercontent.com/assets/24468450/21095602/3812cbf0-c082-11e6-96a7-803eb6656b15.png)

Please let me know if you need more information.

mathiasrw:
Thanks for a good bug... lets see who will get the time to fix it...

kanghj:
The issue seems to be an additional BOM added to the front of the file. Possibly caused by the version upgrade of FileSaver in 0.3.2. I'll try to submit a pull request sometime this week.

mathiasrw:
@junedansari 

Can yuo check if this problem is fixed with the last version of https://github.com/agershun/alasql/blob/develop/dist/alasql.fs.js ?

Dhananjay05414:
Thanks, I have tested and it's working fine now.

junedansari:
@mathiasrw Thank you so much, itz working fine now.

PR #804, opened 2016-12-20
turn off autobom for #793

kanghj:
for #793

We can see that there's a new `disableAutoBOM ` in https://github.com/agershun/alasql/commit/b881f3a7effc217fb316230132f2db4e1b0ed881#diff-dc5eaddfec51693151d584d4cd59b0cf. A quick fix is simply to disable the autobom for csv files, since alasql already handles the bom (with the `utf8Bom` argument for CSV).

Not sure how to test this, since all the tests are in a node environment. Please let me know how to create one if required! 
Currently, a simple test:

```
var data = [["Minsk",100000], ["Riga ë",200000]];

window.exportData = function exportData() {
        alasql("SELECT * INTO CSV('cities.csv', {headers:true, separator:','}) FROM ?",[data]);
}
```
with a button with `onclick="window.exportData()"` can be used to manually check


mathiasrw:
You are awesome - thank you :)

PR #807, opened 2016-12-31
"selectfn" is not declared

nickdeis:
Causes issues when "use strict" is declared


mathiasrw:
Awesome! thanks for fixing this for everybody :)

Issue #818, opened 2017-02-06
minor parameter issue


edwardsmarkf:
hello - 

using objects as  parameters does not appear to work for me.  i attempted to use the syntax found here:

https://github.com/agershun/alasql/wiki/Getting%20started

using this syntax:

     alasql.exec('INSERT INTO one (:a, :b);', {a:3,b:4});

http://jsfiddle.net/mtt9r7rx/7/

mathiasrw:
Crazy - we lost some syntax convenience somehow. 

Its not working because its missing `VALUES` - but the example worked in earlier versions. 

This feature is now part part of the regression tests and is implemented. 

Until next release gets out you can use https://github.com/agershun/alasql/blob/develop/dist/alasql.fs.js

edwardsmarkf:
hello - 

should i be able to use objects in an UPDATE as well?
```
alasql.exec('UPDATE one SET :a WHERE :b',  {a: 987654321, b: 4} ) ;
```

http://jsfiddle.net/mtt9r7rx/16/

mathiasrw:
The changes are now released as part of version 0.3.7

PR #821, opened 2017-02-07
allowing for omitted column names during insert

edwardsmarkf:
Thank you very much for your solution to [#818](https://github.com/agershun/alasql/issues/818) - i asked [StackOverflow ](http://stackoverflow.com/questions/42095671/using-npm-to-install-a-development-version)for suggestions on how to implement: (question upvotes are welcome)
```
npm install 'https://github.com/agershun/alasql#f19e17eeb908638ef320539e5b80357bfcc0709d' ;
```

When a column-name is omitted from the INSERT, it causes issues:
```
alasql.exec('CREATE TABLE one (a string, b string, id integer );');   // notice id integer
alasql.exec('INSERT INTO one (:a, :b  );', {a:1,b:2   });      // :id omitted
```
http://jsfiddle.net/mtt9r7rx/13/

my proposed solution:

"node_modules/alasql/dist/alasql.fs.js" line 12856 

old code:
```
q += values[idx].toJS();
```
new code:
```
if  (  typeof values[idx] !== 'undefined'   )       {       // added by mark
   q += values[idx].toJS();
} else {                                                    // added by mark
   q += null;                                      // added by mark
}                                                           // added by mark
```


mathiasrw:
Hmm... We dont know if the column can represent a null value. 

We have to think more about this. 

Any inputs on how other DB systems solves this?

edwardsmarkf:
In the mariaDB world (mysql/oracle too) it depends on how the column-name is created.  The default is **NULL** - so the following are identical:

```
CREATE TABLE one (a string, b string,       id integer            );
CREATE TABLE one (a string, b string,      id integer NULL   );
```

i ran this on MariaDB as an example (notice the NOT NULL defaults to zero):
```
MariaDB [exampleDb]> CREATE TABLE one (a VARCHAR(50), b VARCHAR(50), id INT );
Query OK, 0 rows affected (0.15 sec)

MariaDB [exampleDb]> INSERT INTO one (a,b) VALUES ( '3', '4' ) ;
Query OK, 1 row affected (0.01 sec)

MariaDB [exampleDb]> SELECT  *  FROM  one;
+------+------+------+
| a    | b    | id   |
+------+------+------+
| 3    | 4    | NULL |
+------+------+------+
1 row in set (0.00 sec)

MariaDB [exampleDb]> CREATE TABLE two (a VARCHAR(50), b VARCHAR(50), id INT **_NOT NULL_** );
Query OK, 0 rows affected (0.06 sec)

MariaDB [exampleDb]> INSERT INTO two (a,b) VALUES ( '3', '4' ) ;
Query OK, 1 row affected, 1 warning (0.04 sec)

MariaDB [exampleDb]> SELECT  *  FROM  two;
+------+------+----+
| a    | b    | id |
+------+------+----+
| 3    | 4    |  0 |
+------+------+----+
1 row in set (0.00 sec)
```
You could specifiy something like this too:
```
MariaDB [exampleDb]> CREATE TABLE three (a VARCHAR(50), b VARCHAR(50), id INT NOT NULL DEFAULT 99 );
Query OK, 0 rows affected (0.04 sec)

MariaDB [exampleDb]> INSERT INTO three (a,b) VALUES ( '3', '4' ) ;
Query OK, 1 row affected (0.00 sec)

MariaDB [exampleDb]> SELECT  *  FROM  three;
+------+------+----+
| a    | b    | id |
+------+------+----+
| 3    | 4    | 99 |
+------+------+----+
1 row in set (0.00 sec)
```

mathiasrw:
Thank you for the investigation @edwardsmarkf

edwardsmarkf:
using cockroachdb, i get the identical results.   

and this is from the[ postgresql INSERT documentation:](https://www.postgresql.org/docs/9.1/static/sql-insert.html)

> Each column not present in the explicit or implicit column list will be filled with a default value, either its declared default value or **_null_** if there is none.

mathiasrw:
Im still thinking about this one. The nature of Javascript lead this library to create this strange duality between null and undefined for results. To change this is not a small effort.

edwardsmarkf:
ok but doing something like this still causes some significant issues:

    alasql.exec('CREATE TABLE one (a string, b string, id integer );');   // notice id integer
    alasql.exec('INSERT INTO one (:a, :b  );', {a:1,b:2   });      // :id has been omitted

result:

> Uncaught TypeError: Cannot read property 'toJS' of undefined

http://jsfiddle.net/mtt9r7rx/13/

mathiasrw:
Hmm. I see. I had a chat with @agershun about this. 

For the moment we need to indicate what fields we are putting data into by adding a `(field1, field2)` after the table name

Example: http://jsfiddle.net/b2519m4y/

If someone will provide code to fix the null vs. undefined vs. default value that would be awesome, but I will close this for now. 

Thank you so much for your persistence and the energy you put into investigating and documenting.

edwardsmarkf:
in the meantime, there is still the issue of alasql crashing if a field is omitted.

what about something like this?

<code>
if  (  typeof values[idx] !== 'undefined'   )       {       // added by mark
   q += values[idx].toJS();
} else {                                                    // added by mark
   q += '""";                                      // add double quotes 
}   
</code>

Issue #824, opened 2017-02-14
Export to xlsx does not work


akob:
Hi!
I want to have multiple worksheets with custom headers. But I have an exception 

**Uncaught TypeError: Cannot convert undefined or null to object**

see example 

https://jsfiddle.net/we884d53/19/

```
window.saveFile = function saveFile () {
var data1 = [{a:1,b:10},{a:2,b:20}];
    var data2 = [{a:100,b:10},{a:200,b:20}];
    var opts = [{sheetid:'One',headers:true},{sheetid:'Two',headers:true}];
    var res = alasql('SELECT a AS Alpha, b AS Betta INTO XLSX("restest344b.xlsx",?) FROM ?',
                     [opts,[data1,data2]]);
}
```

akob:
HI! When it will be fixed?

mathiasrw:
I have no idea. As soon as someone makes a PR ill release a new version.

PR #829, opened 2017-02-19
#824 test & hack for multisheet xlsx export with custom columns

kanghj:
Let's see if this works as a temporary fix for #824. Unfortunately I wasn't able to fix it in a less hackish way. Even if this isn't merged, it can hopefully provide some help to people having the problem described in #824 until a proper fix is found.


mathiasrw:
Looks good to me - AlaSQL is already one big hack on javascript objects... :)

Thank you for taking the time to getting this splinter out of the library !

mathiasrw:
The changes are now released as part of version 0.3.7

PR #841, opened 2017-03-18
Fix #727 - INTO CSV caused regex error with empty quote

### LINKED ISSUE #727: Cant have empty quote in csv
If quote is set to an empty string for `INTO CSV('sdfsfd', {quote:''})` it gives an error because the regexp cant compile

```
/usr/lib/node_modules/alasql/dist/alasql.fs.js:14234
                        s = (s+"").replace(new RegExp('\\'+opt.quote,"g"),'""');
                                           ^
SyntaxError: Invalid regular expression: /\/: \ at end of pattern
    at new RegExp (native)
```

jrgoldfinemiddleton:
This appears to resolve the bug but it sure would be weird to set the `quote` property to `''`.

### DIFF
--- a/src/830into.js
+++ b/src/830into.js
@@ -203,7 +203,8 @@ alasql.into.CSV = function(filename, opts, data, columns, cb) {
 	data.forEach(function(d){
 		s += columns.map(function(col){
 			var s = d[col.columnid];
-			s = (s+"").replace(new RegExp('\\'+opt.quote,"g"),'""');
+                        var quotereg = (opt.quote === '') ? '' : '\\' + opt.quote;
+			s = (s+"").replace(new RegExp(quotereg,"g"),'""');
 //			if((s+"").indexOf(opt.separator) > -1 || (s+"").indexOf(opt.quote) > -1) s = opt.quote + s + opt.quote; 
       
       //Excel 2013 needs quotes around strings - thanks for _not_ complying with RFC for CSV 
--- a/src/830into.js
+++ b/src/830into.js
@@ -203,7 +203,7 @@ alasql.into.CSV = function(filename, opts, data, columns, cb) {
 	data.forEach(function(d){
 		s += columns.map(function(col){
 			var s = d[col.columnid];
-                        var quotereg = (opt.quote === '') ? '' : '\\' + opt.quote;
+			var quotereg = (opt.quote === '') ? '' : '\\' + opt.quote;
 			s = (s+"").replace(new RegExp(quotereg,"g"),'""');
 //			if((s+"").indexOf(opt.separator) > -1 || (s+"").indexOf(opt.quote) > -1) s = opt.quote + s + opt.quote; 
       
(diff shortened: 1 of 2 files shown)


jrgoldfinemiddleton:
When I cloned the repository the "1) Test 140 JavaScript Functions 7. Create table with default conversion Date" was failing.  I believe that might be why the CI build is failing but I'm not sure.  This is one of my first PRs to an open source project.

src/830into.js
```diff
@@ -203,7 +203,8 @@ alasql.into.CSV = function(filename, opts, data, columns, cb) {
 	data.forEach(function(d){
 		s += columns.map(function(col){
 			var s = d[col.columnid];
-			s = (s+"").replace(new RegExp('\\'+opt.quote,"g"),'""');
+			var quotereg = (opt.quote === '') ? '' : '\\' + opt.quote;
```
mathiasrw: Having no quote represent that no replcaement is needed - so i feel it would be better if `opt.quote` was a test condition of if replacement is needed

test/test727.js
```diff
@@ -0,0 +1,18 @@
+if(typeof exports === 'object') {
+	var assert = require("assert");
+	var alasql = require('..');
+	var fs = require("fs");
+}
+
+var test = '727'; // insert test file number
```
mathiasrw: Makes sense that you gave it the ID of the PR - but please give it 612 as thats the next number in the row of test files

test/test727.js
```diff
@@ -0,0 +1,18 @@
...
+describe('Test '+test+' - INTO CSV', function() {
+	
+	after(function(){
+		fs.unlink('sdfsfd.csv');
+        });
+
+	it('With quote = \'\'', function(){
+		alasql("SELECT \"1\" INTO CSV('sdfsfd', {quote:''})");
+        });
+});
```
mathiasrw: There should always be an `assert(...)` in a test. I can see the point in only testing that it does not fail but would be good to read the file (with fs) and verify the content is as you would like it to be. 

What about if we have more than one row? What if we have a string with more words?

Would be good to see som tests verifying a few more scenarios.

mathiasrw:
Hi @jrgoldfinemiddleton

Thank you so much for taking time to fix this!

in https://travis-ci.org/agershun/alasql/jobs/212380956#L782 i looks like test 727 is failing. Did you do a `npm test` on your own computer? It will run the same test as on travis

jrgoldfinemiddleton:
Thank you for the feedback.  I've worked on the changes you suggested and submitted them as a new PR at https://github.com/agershun/alasql/pull/843 .  As for the Travis errors... I don't understand why they are happening.  When I run `npm test` locally the tests pass without error.  During my development process there were errors, but when I made the newest commit to update this PR I had gotten rid of them on my local box.  Could you please check that PR?

mathiasrw:
Sure - ill close this one then (as I understand the new PR has all the same changes - yes?)

jrgoldfinemiddleton:
Correct, all changes are in the new PR.  Thanks for your patient help.

PR #843, opened 2017-03-19
727 into csv travis fix

### LINKED ISSUE #727: Cant have empty quote in csv

jrgoldfinemiddleton:
This PR should supersede https://github.com/agershun/alasql/pull/841 to fix #727  .  The new tests do not cause errors on my local machine.

It seems that with commit [af14329](https://github.com/agershun/alasql/pull/843/commits/af14329f77f8cecb5292f402fe79556cf68d06bb) only the Node.js 4 Travis build is failing for some reason.


jrgoldfinemiddleton:
It appears there are no longer CI errors!  Not sure what changed.

jrgoldfinemiddleton:
This for #727 .

mathiasrw:
The Travis sometimes has a timeout error. I asked to rerun the test and it turned it all green :-)

dist/alasql.js
```diff
@@ -14454,7 +14513,9 @@ alasql.into.CSV = function(filename, opts, data, columns, cb) {
 	data.forEach(function(d){
 		s += columns.map(function(col){
 			var s = d[col.columnid];
-			s = (s+"").replace(new RegExp('\\'+opt.quote,"g"),'""');
+			if (opt.quote !== '') {
+				s = (s+"").replace(new RegExp('\\'+opt.quote,"g"),'""');
```
jrgoldfinemiddleton: It seems to me that we need to clarify what the expected behavior is when `opt.quote` is defined to be a character that appears in the data strings.

For example, if `opt.quote = '?'` and one of the column values is the string `'test?me'`, do we want the output in the CSV file to be `?test""me?`?  If `opt.quote = '"'` and one of the column values is the string `'help"me'`, should the CSV output be `"help""me"`?  This is the current behavior.

With respect to the case where `opt.quote = ''`, the CSV output should always have the format of `"string goes here"`, correct?  This is because we're keeping compatible with Microsoft Excel.

jrgoldfinemiddleton:
It seems to me that we need to clarify what the expected behavior is when `opt.quote` is defined to be a character that appears in the data strings.

For example, if `opt.quote = '?'` and one of the column values is the string `'test?me'`, do we want the output in the CSV file to be `?test""me?`?  If `opt.quote = '"'` and one of the column values is the string `'help"me'`, should the CSV output be `"help""me"`?  This is the current behavior.

With respect to the case where `opt.quote = ''`, the CSV output should always have the format of `"string goes here"`, correct?  This is because we're keeping compatible with Microsoft Excel.

test/test612.js
```diff
@@ -0,0 +1,144 @@
...
+						console.error(err);
+					}
+				});
+			});
+        	});
+	});
+
+	it('With quote = \'\', single multiword string value', function(){
+		alasql("SELECT 'swing out' AS `colname` INTO CSV('test612-1', {quote:''})");
+		fs.open('test612-1.csv', 'r', function(err, fd) {
```
jrgoldfinemiddleton: Thank you for that.  I've made the changes, but I'm having issues with the assertions.  Here's what the first `it()` call looks like for me now:

```javascript
it('With quote = \'\', single string value', function(){
		return alasql.promise("SELECT 'swing' AS `colname` INTO CSV('test612-0', {quote:''})")
			.then(function(state) {
				console.log(fs.readFileSync('test612-0.csv', 'utf8') + '\n\n');
				assert(fs.readFileSync('test612-0.csv', 'utf8') === 'colname\nswing\n');
		}, function(err) {});
	});
```
It appears that the string `readFileSync()` is pulling is not not what I imagine it to be because the assertion fails.  I've also tried the following strings and various other combinations I've forgotten:
`'colname\nswing\n\ufeff'`
`'colname\r\nswing\r\n\ufeff'`
`'colname\nswing\n\n`

I'm at a loss for how to write the assertion. Any ideas on where I might be going wrong with this?

Also, I chose the return the Promise rather than use `done()` because I read that mocha will catch the errors automatically.  That part seems to be working now, whereas I was getting errors using `done()` as you advised.

mathiasrw: Dont worry, we will get there. 

- As the alasql call is async we need to givethe testframework a way of knowing when the call is done. We do that by calling the fist parameter of the callback function to the `it(...` function. Its not named at the moment. Lets call it `done`

````
it('With quote = \'\', single string value', function(done){
```

and then call it after the assert with `done()`

- The `then(` function does not hvae a second parameter, but return a promise where we can `catch` errors.

- Also, reading a file is expensive. If you need the data twice 

- The assert function can take a true/false (you get that by the `===`) but it works better if you give each side as a parameter - because then the test will output what was expected and what we got instead (making it easyer to debug)

In all I suggest the code looks like 

```js
it('With quote = \'\', single string value', function(done ){
	alasql.promise("SELECT 'swing' AS `colname` INTO CSV('test612-0', {quote:''})")
		.then(function(state) {
			var filecontent = fs.readFileSync('test612-0.csv', 'utf8');
			assert(filecontent,'colname\nswing\n');
			done();
		}).catch(err) {
			//console.error(err)
		}
	;
});
``` 

I dont get the `\ufeff` part of the code you are trying to assert

mathiasrw:
Hmm. Its a good question. 

I would say that for 

> For example, if `opt.quote = '?'` and one of the column values is the string `'test?me'`, do we want the output in the CSV file to be `?test""me?`?  

I would expect `?test??me?`

a its the quotechar that must be escaped. Im not super sure about it, so please come with inputs.

jrgoldfinemiddleton:
Hey Mathias, I've figured out how to get the tests and assertions to work correctly now.  I've opted not to use the `done` callback in favor of just returning the promise.  I've tested my test code by intentionally making the assertions fail and believe it is solid.

The `\ufeff`character represents the BOM which is not stripped out from the return value of `readFileSync()`.  Now the tests are passing.  I discovered this by looking at the `Buffer` return type.

And I suppose I see what you're saying regarding escaping the the quote character, whatever it happens to be.  Right now, the code will always replace a quote character in the middle of the field with `""`, but if the quote character is set to `?` we should change the behavior so that it would be replaced with `??`.  I just want to confirm with you that I should go ahead and make that change first.  Anyone else is welcome to chime in.

mathiasrw:
Good work! had no idea I could just return the promise...

> but if the quote character is set to ? we should change the behavior so that it would be replaced with ??

guess you are right.

Ahh, haha, its the bom :) just give the CSV command the option `utf8Bom: false` as described in https://github.com/agershun/alasql/wiki/Csv#options

jrgoldfinemiddleton:
Okay, fix applied to escape all quote characters other than `''`.  Test case added for it as well.  Should be good to merge.

mathiasrw:
Awwweeesommmme

jrgoldfinemiddleton:
Once again, thank you so much for your patient assistance.  You've helped me feel much more confident in contributing to an open source project.

mathiasrw:
:+1: :)

https://twitter.com/rangelwulff/status/844746060622815233

Issue #462, opened 2015-11-19
AUTO_INCREMENT and LocalStorage


error39:
When using localstorage, AUTO_INCREMENT doesn't seem to function, as though without specifying a storage method it does work, is this intended behaviour or am I implementing it incorrect?

my implementation of A_I with localstorage: http://jsfiddle.net/gym7L42r/

mathiasrw:
Nice catch - a clear bug...

Made a version making it easy to compare expected and actual output

http://jsfiddle.net/hfcpcnxt/

error39:
okay, glad I did report it, then, will make alasql a tiny bit better then!
Really appreciate all the work that has gone into this GREAT project, thumbs up!

mathiasrw:
:+1:

glaucocustodio:
Same problem here, it does not generate id...

error39:
I was wondering, will this be fixed?

mathiasrw:
Sorry my short answer from a phone. Yes it will be solved - it's an error - but can't tell you when. In this moment we are transforming the code base to ES6. It depends on when people like @agershun take the time to get into the nitty gritty...

error39:
Okay, I understand the problem.
But do you have a slight idea when this (kind of basic) functionality will be added/fixed then?
In - like - a month or more likely 6 months or a year?
You've all done a great job with this library and for future projects I'd be happy to know if it's actively maintained, or if it's just a more or less finished project?

Wish I could be of some help on the subject, but my Javascript skills are far from suited for these kind of projects.

Good luck with the codebase rewrite, a huge job, I guess..

mathiasrw:
Sorry - I really dont know when someone feel in the mood to dig into issue. 

From November 27 to December 27 we had the following activity: 

- 6 authors have pushed 44 commits (4 Pull requests)
- 86 files have changed 
- 19 Issues created
- 10 Issues closed

So guess some people still seek to improve the library. 

I am sure that if you have an urgent need you can hire @agershun to spend the time needed to fix this.

error39:
Thanks, I'll wait till someone will pick up the issue.
In the mean time I'll stick to a custom workaround which auto-increments at every INSERT statement

agershun:
Sorry for late answer, I will try to fix it in one or two days.

Отправлено с iPhone

error39:
Thanks a lot @agershun!

agershun:
It was harder than expected... Sorry for delay

[...]

PR #847, opened 2017-03-22
Autoincrement for local storage

kanghj:
http://jsfiddle.net/80smosgp/2/ should show that it works as expected. 

This is really @seb-ster 's code provided in the issue thread #462


kanghj:
@mathiasrw, I'll probably need help on the tests. This scenario here only affects localstorage, but all the tests seem to be running in a node environment, how do I test local storage specific code?

mathiasrw:
Of cause...

I suggest you do a 

    npm install --save-dev node-localstorage

and make a test that starts with

````
if (typeof localStorage === "undefined" || localStorage === null) {
  var LocalStorage = require('node-localstorage').LocalStorage;
  localStorage = new LocalStorage('./testxxx.localstorage');
}
```

Then everything should work :)

kanghj:
@mathiasrw, I have modified an existing test to only pass with the modifications added in this PR. Looks like there was support for writing LocalStorage tests in alasql already after all! Very cool. Unfortunately, there are other places where autoincrement for locastorage ought to work, but somehow still isn't (e.g. test389 when AUTOCOMMIT is off), I'll investigate them.

mathiasrw:
Let me know when to merge the PR

mathiasrw:
Any news on this?

kanghj:
Sorry, I got a bit busier than expected and this was a little harder to track down than I thought... Perhaps let's merge this PR first, but not close the issue. I'll work on this sometime next weekend when I have more time.

PR #909, opened 2017-07-20
Case insensitive table/column names

amitauras:
1) yy.casesensitive option doesn't update after changing value of alasql.options.casesensitive. Therefore, used alasql.options.casesensitive in alasqlparser.js

2) Now, as alasqlparser file when case-sensitive options set to false sets all the column names to lowercase therefore in 84form.js file we are changing all column names to lowercase. This way now whatever case whether upper, lower or mixed case column names are specified in excel then also alasql can read data for all those columns when alasql.options.casesensitive is set to false.

I have commited only my changes to the repository. I have not commited changes after build which includes changes in dist folder etc.


mathiasrw:
This looks great - thank you!

Please submit the builded files too so the automated tests can let us merge the code :)

src/84from.js
```diff
@@ -405,9 +405,9 @@ function XLSXLSX(X,filename, opts, cb, idx, query) {
 			var col = alasql.utils.xlsnc(j);
 			if(opt.headers) {
 				if(workbook.Sheets[sheetid][col+""+row0]) {
-					hh[col] = workbook.Sheets[sheetid][col+""+row0].v;
+					hh[col] = alasql.options.casesensitive ? workbook.Sheets[sheetid][col+""+row0].v : (workbook.Sheets[sheetid][col+""+row0].v ? workbook.Sheets[sheetid][col+""+row0].v.toLowerCase() : workbook.Sheets[sheetid][col+""+row0].v);
```
mathiasrw: This works - but its not easy for people to understand. 

Please restructure without using the `if?then:else;` structure.

src/84from.js
```diff
@@ -405,9 +405,9 @@ function XLSXLSX(X,filename, opts, cb, idx, query) {
...
 				} else {
-					hh[col] = col;
+					hh[col] = alasql.options.casesensitive ? col : (col ? col.toLowerCase() : col);
```
mathiasrw: Please restructure without using the `if?then:else;` structure.

test/test614.js
```diff
@@ -0,0 +1,28 @@
+if (typeof exports === 'object') {
+    var assert = require("assert");
+    var alasql = require('..');
+}
+
+var test = '614'; // insert test file number
+
+describe('Test ' + test + ' - Read data from column irrespective of case in query', function () {
```
mathiasrw: Thank you so much for remembering to make a test for your PR :)

test/test614.js
```diff
@@ -0,0 +1,28 @@
...
+        alasql('use test' + test);
+        alasql.options.casesensitive = false;
+    });
+
+    after(function () {
+        alasql('drop database test' + test);
+        alasql.options.casesensitive = true;
+    });
+
+    it('1) Read Data and Check Value', function (done) {
```
mathiasrw: Please describe the test in a way that tells what it is doing. Example could be "Mixed case excel with camel case select"

test/test614.js
```diff
@@ -0,0 +1,28 @@
...
+        alasql('SELECT Account FROM XLSX("' + __dirname + '/test614.xlsx")', [], function (res) {
+            assert.equal(res[0]['Account'.toLowerCase()], 12);
+            done();
+        });
+    });
```
mathiasrw: I suggest having a few more tests both from a positive and a negative side.

A negative test would be that setting the flag to false, the same test would give an error / not find data

there could also be two columns with different case'ing setup (one only upper, one only lower) where it could be tested what happens on casesensetive = off or on

amitauras:
Hi,

I am working on the changes as suggested and I will submit those changes
soon. It's my pleasure that I can be of any help to open source community.
For automated tests to pass I will commit all the files which comes in
source control on build.

Regards
Amit
+91-8285400128

amitauras:
1) removed if?then:else, restructured the code and added comments
2) added test cases for Mixed Case, Upper Case and Lower Case. Both positive and negative tests are used
3) added files generated at build time in the commit

mathiasrw:
Looks really really good. Thank you for taking the time to contribute...

amitauras:
It's my pleasure to have any contribution to such a great library.

Issue #901, opened 2017-07-11
On Passing empty excel file throw error: Cannot read property 'split' of undefined


amitauras:
When we try to upload _an empty excel file_ it throws error in console like following 

```
Uncaught TypeError: Cannot read property 'split' of undefined
    at alasql.js:15106
    at FileReader.reader.onload (alasql.js:3383)
```

Here is a jfiddle link where we can try uploading empty excel file and in the console it thrown error: https://jsfiddle.net/amitauras/3k8j20ds/

Please look into the same and let me know if there is any solution to the same.

mathiasrw:
This is definitely a bug - it should return an empty array. 

While we solve this, i strongly suggest that you use the promise notation - as in this example: https://jsfiddle.net/L52v85m0/

amitauras:
Thank you for the suggestion. Using the promise notation I am able to handle the error. I really appreciate such a quick response.

Rajvardhan222:
hey , is this issue still open i am new to open source and would like to contribute on any issue that would be a good fit for me for starting in this repo.

PR #912, opened 2017-07-21
Read empty excel files without error

amitauras:
1) restructured code in 84from.js so that it can handle empty excel files also.
2) testing both empty and non-empty files in test
3) including files generated at build time also
4) This closes issue #901


mathiasrw:
Sorry for the unpolite response from "codacy" - i have removed the integration until the codebase itself is of a higher quality...

amitauras:
Thank you for the same. I would like to confirm, is something pending from my side that would allow these changes to be able to merge?

PR #921, opened 2017-07-27
Use created database id on foreign key check as default database id

roiemir:

### DIFF
--- a/test/test616.js
+++ b/test/test616.js
@@ -1,0 +1,18 @@
+if (typeof exports === 'object') {
+    var assert = require("assert");
+    var alasql = require('..');
+}
+
+var test = '616'; // insert test file number
+
+describe('Test ' + test + ' - Foreign keys on created database', function () {
+    it('Check foreign key on a created database', function (done) {
+        var db = new alasql.Database();
+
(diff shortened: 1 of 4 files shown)


test/test616.js
```diff
@@ -0,0 +1,18 @@
...
+describe('Test ' + test + ' - Foreign keys on created database', function () {
+    it('Check foreign key on a created database', function (done) {
+        var db = new alasql.Database();
+
+        db.exec("CREATE TABLE Parent( ParentId integer, CONSTRAINT PK_Parent PRIMARY KEY(ParentId) )");
+        db.exec("CREATE TABLE Child( ChildId integer, CONSTRAINT FK_Child_Parent FOREIGN KEY (ChildId) REFERENCES Parent(ParentId) )");
+        db.exec("INSERT INTO Parent(ParentId) VALUES(1)");
+        db.exec("INSERT INTO Child(ChildId) VALUES(1)");
+        done();
+    });
```
mathiasrw: Thank you so much for also including a test. 

We need to include mosre tests demonstrating some aspekcts like 
- Data has been stored in parent 
- Data has been stored in child
- A scenario where the `reference` part is actively demonstrated (normally by doing something that gives one result with the `reference` and another result without the `reference`

roiemir: It's just an exception fix

mathiasrw:
Awesome - thanks :+1:

PR #922, opened 2017-07-28
Remove implicit any in definition file

besquared:
Implicit any causes compiler errors if noImplicitAny flag is enabled.


mathiasrw:
Awesome - thank you!

besquared:
This change seems to have gone missing in the develop branch. What happened here?

mathiasrw:
I was too fast when pulling the PR. The change was made in dist folder thats deleted and rebuild every time.

I have now reimplemented in https://github.com/agershun/alasql/blob/develop/partners/typescript/alasql.d.ts 

Will come out with next release...

PR #929, opened 2017-08-13
Added LTRIM and RTRIM  built-in-defined functions.

jeradgodreault:
## Summary ##
- Added LTRIM and RTRIM built-in-defined functions. Based of the T-SQL documentation. 
- Added unit tests to test these functions.


jeradgodreault:
I'm pretty new to contributing to open source projects. Still trying to get the hang of it!

src/55functions.js
```diff
@@ -157,7 +157,17 @@ stdlib.LOWER = stdlib.LCASE = function(s) {return und(s,'String(y).toLowerCase()
 //stdlib.LCASE = function(s) {return '('+s+').toLowerCase()';}
 
 
-// LTRIM
+// Returns a character expression after it removes leading blanks.
+// see https://docs.microsoft.com/en-us/sql/t-sql/functions/ltrim-transact-sql
+stdlib.LTRIM = function(s) {
+    return und(s, 'y.substr(y.indexOf(y.trim()), y.length)');
```
mathiasrw: it took me a little time to find out what is going on. 

I realise that you are doing a full trim to find the final string to identify the how much to remove. There seem to be some performance issues when dealing with long strings. Also, it feels a bit off to trim both ends to be able to trim one end. 

I suggest you change to a solution based on regular expression. I believe something like the following should do the trick. 

``` 
y.replace(/^\s+/,"");
```

src/55functions.js
```diff
@@ -157,7 +157,17 @@ stdlib.LOWER = stdlib.LCASE = function(s) {return und(s,'String(y).toLowerCase()
...
+// Returns a character expression after it removes leading blanks.
+// see https://docs.microsoft.com/en-us/sql/t-sql/functions/ltrim-transact-sql
+stdlib.LTRIM = function(s) {
+    return und(s, 'y.substr(y.indexOf(y.trim()), y.length)');
+}
+
+// Returns a character string after truncating all trailing spaces.
+// see https://docs.microsoft.com/en-us/sql/t-sql/functions/rtrim-transact-sql
+stdlib.RTRIM = function(s) {
+    return und(s, 'y.substr(0, y.indexOf(y.trim()) + y.trim().length)');
```
mathiasrw: I suggest you change to a solution based on regular expression. I believe something like the following should do the trick. 

``` 
y.replace(/\s+$/,"");
```

test/test617.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+var test = '617'; // insert test file number
+
+describe('Test ' + test + ' - Where LTRIM will remove leading whitespace characters in an expression.', function() {
+
+    it('A) Will remove leading whitespace only', function() {
+        var sql = "select LTRIM('      Hello World !')";
+        var res = alasql(sql);
+        assert.equal(res[0][Object.keys(res[0])[0]], 'Hello World !');
```
mathiasrw: This is had to read for new people coming in. I suggest using `AS ...` to indicate an exact name - or prepend `VALUE OF ` to the `SELECT` to get only the answer

for example: 

```
var sql = "select LTRIM('      Hello World !') AS str";
var res = alasql(sql);
assert.equal(res[0]['src'], 'Hello World !');
```

or something like 

```
var sql = "VALUE OF select LTRIM('      Hello World !')";
assert.equal(alasql(sql), 'Hello World !');
```

mathiasrw:
> First PR and I broken the build D: Not used to this project setup. (sorry!)Do all the changes in the /dist need to be checked in? I notice some of changes hardcode my branch name. Do these get overridden when merged?

I saw you figured out that the dist must be comitted - and yes, all branch specific elements will be dealt with when publishing (its only the version name)

test/test618.js
```diff
@@ -0,0 +1,33 @@
+if (typeof exports === 'object') {
+    var assert = require("assert");
+    var alasql = require('..');
+}
+
+var test = '618'; // insert test file number
+
+describe('Test ' + test + ' - Where RTRIM will remove leading whitespace characters in an expression.', function() {
```
mathiasrw: Can I ask you to add a few tests to see it behaves correctly with tabs and newlines?

jeradgodreault:
Working on all those fixes. Thanks for the review!

test/test618.js
jeradgodreault: Good catch! apparently both tabs and newlines should be ignored. 
Using `/^\s+/` actually includes tabs and whites space characters. So I switched to use `^[ ]+/` to very explicit. 

Also added the unit test as recommended.

src/55functions.js
jeradgodreault: I actually did that first and was really confused why it didn't work. (thinking making regex couldn't be used in the alasql engine).

But after a long time... I figure out the `\s` needed to be escaped to become `\\s`. I guess for these functions there passed through `eval` statement at one point?

src/55functions.js
jeradgodreault: Done

test/test617.js
jeradgodreault: That makes a lot of sense... D'oh!

test/test618.js
jeradgodreault: I ran this query in MS SQL to confirm they were ignored. 

```sql
SELECT 
CASE WHEN RTRIM('Hello' +  CHAR(9) + ' ') = 'Hello' +  CHAR(9) THEN 1 ELSE 0 END as Tabs
, --newline
CASE WHEN RTRIM('Hello' +  CHAR(13) + ' ') = 'Hello' + CHAR(13) THEN 1 ELSE 0 END AS NewLine
```

mathiasrw: I actually was unaware that tabs and newline was not part of the normal SQL timming functionality - thanks for checking up on this

mathiasrw:
Thank you!

Issue #840, opened 2017-03-14
Add ODS support


mathiasrw:
_Spawn from #171_ 

Add ODS from/into

PR #947, opened 2017-10-03
updated xlsx dep, removed xls refs

SheetJSDev:
All of the logic from `js-xls` was folded into `js-xlsx` but some of the browser tests still used very old versions.  Those were cleaned up.

Additionally, `.from.ODS` mirrors `.from.XLSX` (partially addresses #840)


mathiasrw:
Thank you!

Issue #917, opened 2017-07-26
Error in STDDEV/VARP aggregate when aggregating over a zero-length set


alexhisen:
alasql('SELECT STDDEV(col) FROM ?  WHERE 1=0', [[{col:1},{col:2}]]);
will throw an error:
Uncaught TypeError: Cannot read property 'arr' of undefined at Object.E.aggr.VARP

mathiasrw:
Good one !

PR #964, opened 2018-01-10
Fix for aggregate functions applied to empty sets

alexpeattie:
### Current behaviour

When you call most of the built in aggregate functions (`STDEV`, `VAR` etc.) on an empty results set (e.g. `SELECT STDDEV(col) FROM table WHERE 1=0`) an error is thrown, as detailed in #917.

### Expected behaviour

No error should be thrown and `undefined` (or `null`?) should be returned.

### Cause

An aggregate function has 3 stages (as detailed [here](https://github.com/agershun/alasql/wiki/User-Defined-Functions#aggregators)), stage 1 for the first value, stage 2 for each subsequent value, and stage 3 to perform any last transforms on the accumulator. The problem is that with an empty set, we skip straight to stage 3 - but the accumulator will always be `undefined`, because it hasn't been set to anything in stages 1/2 (which won't have run, since there's no first value or subsequent values) which usually leads to errors.

### Fix

I propose we can wrap all the aggregate functions in an outer functions which checks whether the accumulator is `undefined` at the post-processing state and then just short-circuits (by returning `undefined`). An alternative would be to add that condition to all the affected aggregate functions, but figured this approach might be simpler ☺️.

The assumption is that user-defined aggregators would handle this case themselves (might be worth mentioning in the docs?).


mathiasrw:
Thank you!

PR #970, opened 2018-01-17
Fix missing ORDER BY direction when calling toString() on AST

### LINKED ISSUE #965: Converting the syntax tree back to SQL does not convert order by properly
When you convert the alasql AST back into SQL the ORDER BY columns lose their specified sort direction.  See linked jsfiddle: http://jsfiddle.net/nizzlenitz/p1s11tc0/

alexpeattie:
Closes #965 

An ORDER BY statement is a `yy.Expression` which has a `direction` and `expression` property:

screen shot 2018-01-17 at 11 40 43

But `direction` was previously ignored by `yy.Expression.toString()`.


mathiasrw:
Thank you!

Issue #1030, opened 2018-07-18
loadFile during CSV parsing seems to swallow exceptions


elipalmer:
AlaSQL is based on unpaid voluntary work. Thank you for taking the time to make it better. 

Hello, thanks for this excellent library.  I seem to be running into an issue found during testing.  I'm trying to load csv data from a string.  I noticed when a newline is not inside of the string, alasql assumes it is a file that needs to be loaded.  Since the string does not have a .csv member, loadFile throws an exception.  However, that exception seems to be swallowed and written to the console rather than returned as a rejected promise to the calling code.

Here is the jsfiddle: _http://jsfiddle.net/mtt9r7rx/38/_

The fiddle response only prints the version number (I am using v0.4.8),

Running this code locally I get this printed to the console:

_/testcode/ala/node_modules/alasql/dist/alasql.fs.js:3487
AlaSQL v0.4.8
							throw err;
							^

Error: ENOENT: no such file or directory, open 'a, a, a, a.csv'_

Am I using this wrong, or is there any path to getting the exception returned so I can handle this error on my end?

Thanks!

elipalmer:
OK, thanks for taking a look.  I'll try to take a look to see if I can fix it.

elipalmer:
I'm working on a fix for this now.  Just a heads up so I don't step on anyone elses toes looking at working on bugs.

mathiasrw:
This is awesome. Thank you so much. Sorry for the long response time - was on vacation :)

PR #1032, opened 2018-07-23
Alasql 1030 - Make sure errors bubble up to promiseExec

elipalmer:
Corrects issue found in #1030.  I found that the error callback function in loadFile was not being passed in or used.  The exception thrown was being unhandled causing the process to die and the error being sent to the console.

I made a note in the code but will mention it here as well.  The promiseExec function has the (err, data) node callback format standard swapped.  I tried switching them, but that resulted in non errors being sent as errors from elsewhere in the code.  I don't yet have a strong enough understanding of the code base to be comfortable switching them and knowing what else needs to be corrected.

There were a lot of spacing inconsistencies so I ran auto indent to normalize everything in the files I touched.  I also ran eslint --fix on the files I touched.

Running the unit tests against master resulted in this:
 1452 passing (10s)
  376 pending

Running the unit tests against the branch resulted in this:
 1456 passing (9s)
  376 pending

It's not clear to me what the pending tests are, but It seems I didn't break any existing tests.

I did not include the dist/ files as it seems those are generated.


elipalmer:
OK, got CI tests to pass.  I didn't realize older versions of node needed to be supported.  Removed async/await references, this should be ready for review.

Thanks!

src/15utility.js
```diff
@@ -818,29 +822,29 @@ function isIE() {
...
+ @return {integer} hash number
+ */
 
 // FNV-1a inspired hashing
 var hash = (utils.hash = function(str) {
 	var hash = 0x811c9dc5,
 		i = str.length;
 	while (i) {
-		hash = hash ^ str.charCodeAt(--i);
+		hash ^= str.charCodeAt(--i);
```
mathiasrw: Neat

src/40select.js
```diff
@@ -368,9 +368,13 @@ yy.Select.prototype.compile = function(databaseid, params) {
 	// Now, compile all togeather into one function with query object in scope
 	var statement = function(params, cb, oldscope) {
 		query.params = params;
-		var res1 = queryfn(query, oldscope, function(res) {
-			//console.log(res[0].schoolid);
-			//console.log(184,res);
+		// Note the callback function has the data and error reversed due to existing code in promiseExec which has the
```
mathiasrw: Great with a good comment. I made the promiseExec myself - so I'm glad you found this bug!

src/84from.js
```diff
@@ -350,7 +350,7 @@ if(false) {
 		parseText(contents);
 	} else {
 		contents = alasql.utils.autoExtFilename(contents, 'csv', opts);
-		alasql.utils.loadFile(contents, !!cb, parseText);
+		alasql.utils.loadFile(contents, !!cb, parseText, query.cb);
```
mathiasrw: Neat.

test/test624CSVParsing.js
```diff
@@ -0,0 +1,56 @@
+if (typeof exports === 'object') {
```
mathiasrw: Thank you for making a good test file too.

elipalmer:
Awesome, thanks for merging!  Glad you had a successful "no computer" vacation, I haven't had one of those in years.

PR #1033, opened 2018-07-26
Alasql 789 -Error dbprecision is not defined while doing a select query

kylebernhardy:
Fixed the scope for assigning dbprecision on parsing the convert statement so that an exception isn't thrown.

Updated the function to make sure dbprecision is passed in on the args.

Coalesced NUMBER, FLOAT, DECIMAL & NUMERIC together to have shared logic.

Changed the existing math to utilize the inherent Math functions which allow for defining precision & scale.

Created new unit test file /tests/test789.js.

Looking forward to feedback on this PR!


kylebernhardy:
It looks like the CI failed due to a bad test unrelated to my fix.  Please advise if this is an issue.

test/test176.js
```diff
@@ -102,8 +102,9 @@ describe('Test 176 - CSV and TSV', function() {
 		) {
 			assert(res.length == 4);
 			//	    console.log(res);
+      done();
 		});
 		//assert.deepEqual(res[1],{ 'Country':'Kazakhstan', 'City':'Almaty' });
-		done();
+
```
mathiasrw: Good catch.

test/test789.js
```diff
@@ -0,0 +1,44 @@
+if (typeof exports === 'object') {
```
mathiasrw: Thank you for providing a dedicated test file.

PR #1036, opened 2018-08-01
Correct grammar of "table/column does not exist" error

zichongkao:
Not sure if I should have left out the changes to dist/ in my commit. 

This change is relatively superficial. No additional tests were required, though I had to edit some existing ones.


zichongkao:
Abandoned this one because there were too many changes to the dist folder. Likely because I'm on Windows and the line endings were messed up.

PR #1040, opened 2018-08-03
Edit grammar of missing table/column error message

zichongkao:
Some of the dist files changed because they were using different
line endings. I standardized them to LF.


mathiasrw:
How awesome that you took the time to fix this. Thank you :)

PR #1045, opened 2018-08-09
Implement TIMESTAMPDIFF function from MYSQL

Vector919:
This PR implements the TIMESTAMPDIFF function for mysql:
https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_timestampdiff

Allowing alasql to better support mysql specific syntax and queries.

Functionality is mostly identical to the DATEDIFF function, but the syntax is slightly different:

```sql
SELECT TIMESTAMPDIFF(MONTH,'2003-02-01','2003-05-01');
```

MONTH, DAY and YEAR are not string params passed to the function in mysql, unlike DATEDIFF.

the parser (alasqlparser.jison) has been updated to account for this case.


mathiasrw:
Not sure why the test is failing. Code looks like it would not affect this. 

Did you do a `npm test` before pushing your last change? did it went ok?

Vector919:
@mathiasrw Yes, it looks like those tests are failing due to a mismatch between the code in 
alasqlparser.jison and the compiled jison in alasqlparser.js

Is the correct way to generate that file still:
```
npm run jison
```
?

mathiasrw:
Yes - first do a `npm run jison` and then a `npm test`

Vector919:
@mathiasrw so it looks like, without any changes, running `npm run jison` and then `npm test` has these same test failures.

mathiasrw:
alasqlparser.js is autogenerated so any changes will be overwritten. 

I will have to look into the issue.

PR #1050, opened 2018-09-01
fix for creating xlsxml ouput with multiple sheets

generia:
Thanks for the great library. Here is a fix for creating xlsxml with multiple sheets.


mathiasrw:
Thank you!

Issue #1008, opened 2018-06-02
XLSX Specified sheet without knowing the name


cscan:
Suggest to add an option to allow to specify the sheet by its position (e.g. 0, 1, 2 ...) instead of using the actual sheet name. Something like this:

alasql(`
select * from XSLX('abc.xlsx', {sheet: 2})
`);

PR #1052, opened 2018-09-05
XLSX Specified sheet without knowing the name #1008

akumadare:
This PR enables a number to be passed in as the sheetid. This will allow the sheet to be referenced by its position instead of name (as per enhancement request #1008 )

Sample usage as follows:

```
alasql.promise('select * from xlsx("test625.xlsx",{sheetid:0})')
.then(function(data){
     console.log(data);
}).catch(function(err){
     console.log('Error:', err);
});
```


mathiasrw:
Such a small change with such a large impact. I love it.

PR #1092, opened 2019-02-25
Fix JOIN CSV(filename)

bjouhier:
This PR fixes CSV functions inside JOIN clauses. 

See #741 which was closed but with a workaround, not a complete fix.

Note: I applied the fix to `alasqlparser.jison` and then manually to `alasqlparser.js` instead of recompiling  `alasqlparser.jison`. If I recompile this file, the commented code at  https://github.com/agershun/alasql/blob/develop/src/alasqlparser.js#L477 gets re-enabled and then tests 394 and 400 fail.


bjouhier:
I previously committed the `dist/alasql.d.ts` file which had been updated by the gulp task. I reverted this file because the new version gives a TS compilation error: `./../node_modules/alasql/dist/alasql.d.ts:83:5 - error TS2666: Exports and export assignments are not permitted in module augmentations.`.
Looks like `partners/alasql.d.ts` needs a fix.

mathiasrw:
You are awesome. Thank yo so much.

PR #1093, opened 2019-02-26
fixed duplicate key + engines syntax error in package.json

bjouhier:
My editor was complaining about two errors in package.json:

* `typings` key was duplicated.
* `engines` key had an invalid value: array instead of object. As I don't know what the minimum supported node version is, I put `"*"`.


mathiasrw:
Thank you!

PR #1094, opened 2019-02-28
Fix source files to get safe build

bjouhier:
This small PR fixes `partners/typescript/alasql.d.ts` and `src/alasqlparser.jison` so that the `dist` files don't get broken when the project is rebuilt.

I rebuilt the project and I included the new `dist` files in this PR so that future PRs only show relevant diffs on `dist` files.

Notes:
* Running the tests modifies the `dist` files because of the sha1 which is included in the files. This is a bit annoying. What about including only the branch name in the version?
* The gulp task fails with a node error (Assertion args[1]->IsString() failed) on node 10.15.1 (darwin). It works fine on node 8.


mathiasrw:
@bjouhier Any inputs to the questions in the review?

I guess the real fix is to not include dist in the repo...

mathiasrw:
@bjouhier any inputs on the questions in the review?

mathiasrw:
Please open a new PR if this is still relevant

PR #1101, opened 2019-04-17
Fixes for "SELECT * ... OUTER JOIN ..."

akhaneev:
There was an issue when when performing OUTER JOIN: see test800.js for input details. The output was with missing id: 4
       [
            {id: '1', a: 'one', b: 'A'},
            {id: '2', a: 'two', b: 'B'},
            {id: undefined, a: 'four', b: undefined},
            {id: '3', a: undefined, b: 'C'},
        ];

Changes are only in src/424select.js + new test800.js
Other files - something about line breaks. Occurred automatically after cloning the repository.


mathiasrw:
Thank your!

PR #1105, opened 2019-05-02
Fix for issue with a chain of OUTER JOINs

akhaneev:

Example:
SELECT * FROM ? T1 ' +
            'OUTER JOIN ? T2 ON T1.id = T2.id ' +
            'OUTER JOIN ? T3 ON T1.id = T3.id OR T2.id = T3.id ' +
            'OUTER JOIN ? T4 ON T1.id = T4.id OR T2.id = T4.id OR T3.id = T4.id

Expected: rows from all tables are joined.
Actual: only all rows from T1 and T2 are present; rows from T3 and T4 that do not match T1 and T2 are missing.

See test801.js


mathiasrw:
You are awesome! Thank you so much. I am very sorry it took me so long to see your PR. Big thank you for digging into this ooooold old old issue. 🎉

mathiasrw:
I will release a new version in the weekend with this.

akhaneev:
Thank you!

Issue #1113, opened 2019-06-26
Drop trigger throws error


umasudhan:
The following program creates a table and a trigger on the table. An error is thrown when the trigger is dropped.

```
const alasql = require('alasql');
alasql.fn.onchange = function(r) {
    console.log('got event in trigger callback 1');
};
alasql('CREATE TABLE one (a INT)');
alasql('CREATE TRIGGER myTrigger AFTER INSERT ON one CALL onchange()');
alasql('INSERT INTO one VALUES (123)');  // This will fire onchange()
alasql('DROP TRIGGER myTrigger');
alasql('INSERT INTO one VALUES (231)');  // This should not fire onchange()
```

```
**Stacktrace:**
got event in trigger callback 1
<path>\node_modules\alasql\dist\alasql.fs.js:14848
		delete db.tables[tableid].beforeinsert[triggerid];
		                          ^
TypeError: Cannot read property 'beforeinsert' of undefined
    at yy.DropTrigger.execute (<path>\node_modules\alasql\dist\alasql.fs.js:14848:29)
    at Function.alasql.dexec (<path>\node_modules\alasql\dist\alasql.fs.js:4659:46)
    at Function.alasql.exec (<path>\node_modules\alasql\dist\alasql.fs.js:4607:17)
    at alasql (<path>\node_modules\alasql\dist\alasql.fs.js:138:16)
    at Object.<anonymous> (<path>\test9.js:8:1)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)

Process finished with exit code 1
```

mathiasrw:
So sorry to see you have to do things like https://github.com/Gippsman2017/node-red-contrib-redlink/commit/47bd90e197787812413e60a2b975f8c19ebf3c2d#diff-ef6ff4525efc31a46136931fbca91995R144 to work around this. 

This would be a great first bug to look at for anyone.

garynewelluk:
Drop trigger throws error #1113

I have put a PR up for this issue. 

The drop was trying to use the trigger as the table id and because that didn't exist it threw the error.

[...]

PR #1140, opened 2019-10-13
Drop trigger throws error #1113 - added table id to the trigger and u…

garynewelluk:
…sed this when dropping the trigger to remove the trigger from the tables


mathiasrw:
@garynewelluk Thank you!

mathiasrw:
Fix #1113

Issue #104, opened 2015-03-25
LEAST and GREATEST : TypeError: Object #<Object> has no method 'join'


gregaubert:
When trying to use the __GREATEST__ or __LEAST__ functions with this exemple :

```javascript
var alasql = require('alasql');
var tab = [{a: 1, b:4}, {a:5, b:3}];
alasql('SELECT GREATEST(a, b) as test FROM ?', tab);
```
We get the error _<Object> has no method 'join' _ because we try to use join on the [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) object, and it's not real array, so it got no method __join__.

It is possible to solve it by calling the __join__ method like this : 

```javascript
Array.prototype.join.call(arguments, ',')
```

But if I update alasql with this fix, then I get __NaN__ results instead of the numbers : 

```javascript
 alasql('SELECT GREATEST(a, b) as test FROM ?', tab); 
// result : [ { test: NaN }, { test: NaN } ]
```

And the result of the __join__ method called on the arguments in the __GREATEST__ method gives this : 
```javascript
console.log(Array.prototype.join.call(arguments, ','));
// result : p['default']['a'],p['default']['b']
```
Is this correct ? Or could the problem come from that ?

agershun:
Mea culpa!

You can temporary fix it with:
```js
    alasql.stdlib.GREATEST = function(){
          return 'Math.max('+Array.prototype.slice.call(arguments).join(',')+')'
    };
    alasql.stdlib.LEAST = function(){
          return 'Math.min('+Array.prototype.slice.call(arguments).join(',')+')'
    };
```
See full example [in jsFiddle](http://jsfiddle.net/mrweh7wo/1/)

I will fix this bug and update the library tonight.

agershun:
BTW: You missed brackets in ```alasql()``` data argument:
```js
    alasql('SELECT GREATEST(a, b) as test FROM ?', tab); // wring
    alasql('SELECT GREATEST(a, b) as test FROM ?', [tab]); // correct
```

gregaubert:
Oh yeah so that's the reason for __NaN__ value. Okay my bad.

For the temporary fix, you don't need to _slice_ before the _join_, this works too : 

```javascript
    alasql.stdlib.GREATEST = function(){
          return 'Math.max('+Array.prototype.join.call(arguments, ',')+')'
    };
    alasql.stdlib.LEAST = function(){
[...]

PR #1156, opened 2020-01-16
fix QUART/MEDIAN/MIN/MAX on number/date/string

### LINKED ISSUE #411: MAX/MIN function not working for words
Apparently, using JSON table data, and aggregating the entire table into one record (no GROUP BY):
the SUM function works fine, but
the MAX/MIN functions return "undefined"
and the GREATEST/LEAST functions return 0.

Here is a sample of the data I use:

```
[ {field1: value1, field2:value2} ],
[ {fieldA: valueA, fieldB: valueB} ]
```

Am I doing something wrong or are the functions incomplete?

Thanks,
Timm

forestfang-stripe:
Fix MIN/MAX and GREATEST/LEAST to work on date and string by actively compare with `<` and `>` instead of `Math.min` and `Math.max`. This is commonly supported by other SQL engine: https://www.w3resource.com/sql/aggregate-functions/min-function.php
Currently, `MIN/MAX` works for Date but the data type changes to number, which is not what user would expect.

Fix QUART/MEDIAN to use custom comparator with `<` and `>`. When comparator is not supplied, Javascript coerce arguments to string:
>If omitted, the array elements are converted to strings, then sorted according to each character's Unicode code point value.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Enabled skipped tests on these functions and update tests for new and correct behaviors (for example, median of numbers between 1 and 10000 should be 5000 not 5499.) add test from #104

p.s. some files got auto-formatted, specifically for `423groupby.js`, I only changed 'MIN' and 'MAX' implementation.

closes #411

### DIFF
--- a/src/423groupby.js
+++ b/src/423groupby.js
@@ -52,7 +52,7 @@ if(false) {
 //			console.log(colid,'ok');
 		} else {
 //			if(colid.indexOf())
-//			console.log(colid,'bad');	
+//			console.log(colid,'bad');
 			var tmpid = 'default';
 			if(query.sources.length > 0) tmpid = query.sources[0].alias;
 //			console.log(new yy.Column({columnid:colid}).toJS('p',query.sources[0].alias));
@@ -126,12 +126,7 @@ if(false) {
 			aft2 = '';
 
 		if (typeof query.groupStar !== 'undefined') {
-			aft2 +=
-				"for(var f in p['" +
-				query.groupStar +
-				"']) {g[f]=p['" +
-				query.groupStar +
-				"'][f];};";
+			aft2 += "for(var f in p['" + query.groupStar + "']) {g[f]=p['" + query.groupStar + "'][f];};";
 		}
 
 		/*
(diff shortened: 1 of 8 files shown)


src/423groupby.js
```diff
@@ -319,41 +300,17 @@ if(false) {
...
-							pre +
-							"g['" +
-							colas +
-							"']=Math.max(g['" +
-							colas +
-							"']," +
-							colexp +
-							');' +
-							post
+							pre + 'if ((y=' + colexp + ") > g['" + colas + "']) g['" + colas + "'] = y;" + post
```
forestfang-stripe: I used `y` here assuming it is intended to be a temporary variable (as far as I can tell from how it is used in `AVG`)

mathiasrw:
You are soooo deep into the details. Wow. I love it. even a nice test case. Perfect.

mathiasrw:
Released as version 0.5.4 💟

PR #1157, opened 2020-01-16
Fix tests for node 12+

### LINKED ISSUE #1135: Node 12
One test is failing in node 12 

https://travis-ci.org/agershun/alasql/jobs/585543277

forestfang-stripe:
fix #1135 

As can be seen in the following snippets, deepEqual behavior has changed in node 12 where it became stricter. Update the test cases to accurately reflect the current behavior. We can discuss whether alasql should return array of arrays in these situations separately. 

https://runkit.com/forestfang-stripe/deep-equal-array/1.0.0
https://runkit.com/forestfang-stripe/deep-equal-array/2.0.0

### DIFF
--- a/test/test046.js
+++ b/test/test046.js
@@ -37,18 +37,31 @@ describe('Test 046', function() {
 		});
 
 		it('queryArrayOfArrays()', function(done) {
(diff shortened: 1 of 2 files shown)


mathiasrw:
THANK YOU!

PR #1158, opened 2020-01-16
Adjust node versions in travis

forestfang-stripe:
both node 8 and 11 are no longer maintained as of 2020/1/1 and node 8 also fails to build due to the fact that one of the dependencies now requires node >=10. update list of node to test against

https://nodejs.org/en/about/releases/


mathiasrw:
Ohh - nice. Thank you so much for taking the time to just look into this :-D

PR #1160, opened 2020-01-17
Updated Sort Function will fix #1159

### LINKED ISSUE #1159: Quartile function incorrect because of sort
Hello, I was using the quartile function in some sql code and noticed it wasnt correct. Using the function definition from src/55functions.js I derived a counterexample:

```
QUART = function(v, s, stage, nth) {
	//Quartile (first quartile per default or input param)
	if (stage === 2) {
		if (v !== null) {
			s.push(v);
		}
		return s;
	} else if (stage === 1) {
		if (v === null) {
			return [];
		}
		return [v];
	} else {
		if (!s.length) {
			return s;
		}

		nth = !nth ? 1 : nth;
        document.write("nth: " + nth + "<BR>");
		var r = s.sort();
        document.write(r);
		var p = (nth * (r.length + 1)) / 4;
		if (Number.isInteger(p)) {
			return r[p - 1]; //Integer value
		}
		return r[Math.floor(p)]; //Math.ceil -1 or Math.floor
	}
};

var i = 0;
var arr = [20, 0.02, 20.02, 896.12, 1510.9, 20.02];
res = 0;
res = QUART(arr[0], null, 1, 1);
for(i = 1; i < 5; i++)
{
	res = QUART(arr[i], res, 2, 1);
}
document.write(QUART(arr[5], res, 3, 1));
```
The problem is that s.sort() uses lexicographical sorting, so it wasn't caught as incorrect in the tests since you only tested using single digit numbers (a perfectly reasonable mistake :) ). All that needs to be changed is to put
```
s.sort(function(a,b){return a - b})
```

aried02:
Change sort from (default) lexicographical sort to a numerical sort.
This resolves #1159

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -330,7 +330,7 @@ alasql.aggr.QUART = function(v, s, stage, nth) {
 		}
 
 		nth = !nth ? 1 : nth;
-		var r = s.sort();
+		var r = s.sort(function(a,b){return a - b});
 		var p = (nth * (r.length + 1)) / 4;
 		if (Number.isInteger(p)) {
 			return r[p - 1]; //Integer value


mathiasrw:
Thank you so much for digging into this :-D

PR #1167, opened 2020-02-16
Adds handling for groupBy within nested array data set/params

sljohnsondev:
Right now, when passing in data params to `alasql.promise()` as nested arrays, the groupBy logic is not behaving as expected b/c the columnid is used to compile the data which breaks when there are multiple tables with the same index value (i.e. `columnid`) in the group by.

To resolve this, I am adding the `nick` value to the group values immediately after they are set to to their respective columns in `424select.js`.  This allows us to use the same nick value when building up the groupby logic/functions in `43rollup.js`.

I also added `test/test803.js` to ensure the updated logic is working as expected.


sljohnsondev:
@mathiasrw any thoughts on when this change might get pushed in a release?  We are eager to get this working in our project for our upcoming release in a few weeks.

sljohnsondev:
@mathiasrw just wanted to follow-up on this to see when the next release might be on this.

mathiasrw:
@sljohnsondev Thank you for poking me. So sorry for the unnecessary wait time. It's up now as v0.5.6 :)

sljohnsondev:
No worries at all, Mathias!  I appreciate the follow-up and hope you are staying safe and healthy where ever you are.

mathiasrw:
Thanks Sam. Sydney is not as bad as many other placers. Are you affected much? Is everything locking down in Denver?

sljohnsondev:
Happy to hear that.  Denver started a full lock down of all non-essential services yesterday but cases seem to be relatively average compared to other parts of the country like NYC and Washington.  Stay safe and sane!

PR #1168, opened 2020-02-21
Fix db use method

CoreyBuckley:
While playing around with alasql and looking into the code, I noticed that if you call alasql.use on a DB that doesn't exist, alasql.useid will always be that DB name. Invalidating any future DB operations if you don't correctly set the useid.

I believe it makes more sense to check if the DB name is valid, _then_ set alasql.useid. As opposed to just setting alasql.useid regardless of its validity.

This PR aims to fix this issue. I've added a corresponding test. :smile:


mathiasrw:
You even added a test - Thank you:)

PR #1175, opened 2020-03-25
Fix usage of CURRENT_TIMESTAMP (Fix #1174)

### LINKED ISSUE #1174: yy.FuncValue toString prototype not handling CURRENT_TIMESTAMP variable correctly
For yy.FuncValue = { funcid: "CURRENT_TIMESTAMP }, the toString() prototype is returning `CURRENT_TIMESTAMP()` which correctly causes an error when passed into alasql.  In order to fix, there needs to be a check for this funcid and skip the step where the () and arguments are added to the string return value.

**Specific Example:**
const foo = new alasql.yy.FuncValue { funcid: "CURRENT_TIMESTAMP };
const result = foo.toString()

_Current behavior - result = `CURRENT_TIMESTAMP()`_
**Expected behavior - result = `CURRENT_TIMESTAMP`**

sljohnsondev:
This PR resolves the issue noted here - https://github.com/agershun/alasql/issues/1174

Added unit tests and confirmed all tests are passing.


mathiasrw:
Thank you!

Released as v0.5.8

Issue #1176, opened 2020-03-31
INSERT's toString() returns the values without parenthesis


ArianJM:
Apparently, the `INSERT` statement's `toString()` doesn't return the values inserted in parenthesis.
Ex:
```javascript
var insertStatementString = "INSERT INTO cities VALUES ('Rome', 2863223), ('Paris', 2249975), ('Berlin', 3517424), ('Madrid', 3041579)"
var statements = alasql.parse(insertStatementString).statements;

console.log(statements.toString());
```
Outputs:
```sql
INSERT INTO cities VALUES 'Rome',2863223,'Paris',2249975,'Berlin',3517424,'Madrid',3041579
```
I'd expect the code to output 
```sql
INSERT INTO cities VALUES ('Rome',2863223),('Paris',2249975),('Berlin',3517424),('Madrid',3041579)
```

jsfiddle: http://jsfiddle.net/6ha2397r/1/

PR #1177, opened 2020-03-31
Update Insert's toString()

ArianJM:
Fixes: #1176 
Quick and small fix. I'm not sure if I'm missing something, hopefully not.

This is my first PR, I'm not sure if I have to do something differently.


mathiasrw:
It looks perfect - and there is even a test. Thank you!

PR #1179, opened 2020-04-15
Composite foreign keys implementation 

kiranramnath:
Hi, 

We will be needing composite foreign keys implemented for a project that is in progress. We did a fix in the handling of foreign key constraints on table creation which now includes composite foreign keys. 

Prettier muddied the code but the update is in 60createtable.js line 281 to 317.

And we have written unit tests in test807.js.

Thanks,
Kiran

FYI : @bjouhier @janwvjaarsveld


kiranramnath:
FYI @mathiasrw , I added a small change, as I was not accounting for null foreign keys

mathiasrw:
Looks great! Thank you!!!

Issue #1148, opened 2019-12-19
UNIQUE with FILESTORAGE not Working!


xxsoadxx:
When i set attach filestorage to UNIQUE example stop working

Expected result: Error: Cannot insert record, because it already exists in unique index
Actual Result: All row inserted 

Code:
        ATTACH FILESTORAGE DATABASE testdb("./testdb.json");
        USE testdb;

        CREATE TABLE usersTable (id INT UNIQUE, name NVARCHAR(255))
        INSERT INTO usersTable VALUES (1,"Andrey")
        INSERT INTO usersTable VALUES (2,"Kate")
        INSERT INTO usersTable VALUES (3,"Kate")
        INSERT INTO usersTable VALUES (3,"Andrey")

PR #1184, opened 2020-04-26
Fix DROP Filestorage Database with database name

RichyRaj:
Hello,

Awesome work with the library. It's amazing to see that we could have all the power of a SQL database inside the browser.  Great work!

I was working on #1148 . While writing tests for the issue, I realized that we cannot DROP a Filestorage database by specifying its database name. I thought I will try to fix this first before moving on to #1148 .
### Problem: 
We are not able to DROP a Filestorage Database by specifying its name. 
For example, running the query:
```sql
DROP FILESTORAGE DATABASE testDBFS
```
would result in the following error: 
`"Error: Cannot drop database file, because it does not exist".`

### Reason: 
The reason this is happening is that FS.dropDatabase (**_94filestorage.js_**) is expecting the file name to be provided when dropping a database. So providing the JSON file path instead of the database name works. However, its method signature and the documentation for Filestorage suggests that we drop the database by specifying its database name. But, when we specify the database name in the drop query, as shown above, it treats the database name as the file name and checks if that JSON file exists. Since there will be no such file, it throws the error.  

This PR adds support to DROP Filestorage databases with the database name specified. The existing behavior of dropping Filestorage databases specifying the filename instead of the database name will still continue to work.

The tests for this fix are in the **_test808.js_** file. 
I have also added comments in the code explaining the issue and the solution wherever there are changes. 

**P.S.** I am currently working on #1148 . I will submit a PR for that as soon as possible.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v0.5.9

PR #1188, opened 2020-05-05
filestorage | join 

tran4o:
1. fix save file on commit with autocommit false 

2. improve left join / join on .. optimizations for complex and on contions

ENABLED BY : 
alasql.options.simplifyJoins=true
				
IMPLIED REQUIREMENT : all join tables MUST have primary key id

TODO ADD search for primary key (unique column ?)

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -273,7 +273,8 @@ stdlib.UPPER = stdlib.UCASE = function (s) {
 // Concatination of strings
 stdfn.CONCAT_WS = function () {
 	var args = Array.prototype.slice.call(arguments);
-	return args.slice(1, args.length).join(args[0]);
+	args = args.filter((x) => !(x === null || typeof x === 'undefined'));
+	return args.slice(1, args.length).join(args[0] || '');
 };
 
 //stdlib.UCASE = function(s) {return '('+s+').toUpperCase()';}
--- a/test/test413.js
+++ b/test/test413.js
@@ -66,7 +66,7 @@ describe('Test 413 CONCAT_WS (issue #429)', function () {
 
 	it('3. Test CONCAT_WS', function (done) {
 		var res = alasql(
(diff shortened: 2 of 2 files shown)


mathiasrw:
@tran4o Interesting!

I have not looked in details in the code, but a few questions:

- Is this _only_ applicable to file storage tables? If so, I feel the name of the config option should be different. 

- Any chance you can add a test file in the /test folder to make sure we dont break this feature if we one day change the code? I suggest making a negative test testing that it throws an error if you are using simplifyJoins

tran4o:
the file storage fix applies only the file storage. it has nothing to to with join optimizations. they are a separate thing.
its just a proposal, 
i don't have any time to write test now, maybe after finishing my investigation about real world usage with enterprise data. I'm using the library as offline storage for real meta-driven  enterprise framework, so all possible real problems are bound to happen. I am almost at final success?,  so maybe in the some feature, when it all clear if the library is worth for my case i can do some test etc. somehow basic stuff in the documentation is impossible to find and the community here does not seem to be interested resolving basic issues.. so i gone resolving them for my self and i can at least "report" my way of hacking them.

I have another "hacked feature" (maybe applies only for filesystem?) : emulating transaction rollback without loading the database again from the storage. (temporary transactions without commit). But it now still bound with provided features from my application layer, so if i manage to make it as  general solution i'll post it as pull request.

mathiasrw:
Do you have any performance measures to support that simplifyJoins are better?

tran4o:
sorry i was not active in this subject long time ago, as much i can still remember without the simplifyJoins hack the queries generated by the our ORM the queries ran extreme long. The ORM queries are quite trivial, but all data is filtered by inheritance with <column> IN (...) and hierarchical access group condition like <column> IN (...) OR <column> IN (...)..  All tables have about 10-50k rows.  
I am also considering duckdb (wasm) in my further investigations for embedded storage, i think we will alasql and duckdb storage for comparison.

mathiasrw:
I think wasm is the right way to go with data like that.

PR #1195, opened 2020-05-19
convertData option

akhaneev:
A new option convertData added - a user-defined function that modifies data before performing operations on it.
My use case: alasql is used to process HTML tables created by people in the rich text editor. Each cell has its text / numeric / date value that alasql can work with and HTML representation like formatting, links, etc. So each cell should preserve its formatting after manipulating tables (for example, joining two tables). The idea is to operate objects with value and view attributes: the value is for SQL processing, the view is for outputting the correctly formatted cell.
How it works. By default the convertData is a simple function that does nothing:
`alasql.options.convertData = function (d) { return d; };`
In my use case this function is replaced with another one, that checks if the data is an instance of my class and extracts the object value that can be compared, calculated and so on, otherwise it returns the data as is:
`      alasql.options.convertData = function (d) {
            return "('object' === typeof (" + d + ") && (" + d +").constructor.name === 'Pair' ? (" + d + ").valueOf() : " + d + ")";
        };`
Hope this feature will be useful for other developers.


mathiasrw:
Hi @akhaneev 

Thank you for making a great PR. I have solved the merge conflicts. 

Did I understand the key concept correct: Its tranforming all data when selecting because when selecting from htmltable you cant manage the input?

Hmm. Doing the tranformation on each output seems counterintuitive. Dont you think it makes more sense to make a function that transforms the data before its loaded?

akhaneev:
Hi @mathiasrw,
Thank you for your time looking into this PR!
The idea is to have the correct view representation for each table value after transforming the table with alasql. A user writes his own query to transform somehow his tables. You can see some use cases [here](https://docs.stiltsoft.com/display/TFAC/Custom+Transformation+-+use+cases+with+advanced+SQL+queries).
Without this feature, I have to get plain text from table cells so that these values can be correctly compared and modified. Then the result is outputted as a new table, but all cells contain plain text.
What I need to have is an object with value for comparison and modification and its view representation to output the resulting table.
I used the idea of the alasql.options.valueof - that was exactly what I needed, but it works in the ORDER function only. And I need the valueOf the object each time it is compared or an operation on this object is performed. So I added the convertData function that is executed when a value is compared or modified. But I didn't find where functions like SUBSTR are executed, so it will fail in this case (I use js Proxy as a workaround).

mathiasrw:
I see. 

Would it not work better to be able to transform data when you read from the table?

What is you read more than one table? The function is applied to all tables. Maybe a function per table would make more sense?

akhaneev:
Not sure I understood your suggestion correctly. If the data is transformed on the table reading, then the resulting table JSON will have only values without a linked view representation. The view-value object has to go through all transformations to keep consistency. The object value is extracted from the object for comparison with another value, but the same object still returned in the alasql result.
For example, we have these values (in pseudo markup):
<url1>A</url1>, <red>B</red>, <green>A</green>, B
We need to execute some SQL that somehow shuffles data, compares, etc. So we feed alasql with values that can be compared (and a user expects those values):
A, B, A, B
And we get the result:
A, A, B
We need to know what of these values is a link or a colored text. 
Operating view-value objects solves this problem. We feed alasql with objects like
{value: A, view: "<url1>A</url1>"}
When values are compared
{value: A, view: "<url1>A</url1>"} = {value: A, view: "<green>A</green>"}
convertData extracts values for comparison and it gets equal to
A = A
And the result looks like 
{value: A, view: "<url1>A</url1>"}, ...
So now we can output the valid HTML view.

If you have any ideas, how it can be accomplished better, an example will help me to understand it better.
Thanks!

akhaneev:
Hi @mathiasrw,
Please, let me know your thought on this PR.

mathiasrw:
Hi @akhaneev 

I am still trying to fingure out if the usecase is worth the added complexity. Maybe its because I dont know a real live example of where I would use this. 

I still have a feeling that it would be better to change data on the way IN than to edit et on each and every select.

akhaneev:
Hi @mathiasrw 
Maybe we could have a call in Zoom where I could show the use case and we could discuss it? You can select time slot [here](https://calendly.com/akhaneev/table-filter-and-charts).

mathiasrw:
@akhaneev Sure. Neat booking thing. I dont use zoom. If we can use google meetup or whereby.com that would be great.

akhaneev:
@mathiasrw, thank you for scheduling the meeting. Let's use Google Meetup then.

akhaneev:
@mathiasrw,
Here is the link meet.google.com/cji-ebek-pdr

mathiasrw:
Looking into a suggestion to use an added prototypr to the needed value type that is added in the modify part of the code so that we channge as little as possible but keep your case flowing.

akhaneev:
Thank you very much! Just let me know if I can help you somehow.

mathiasrw:
Have a look at http://jsfiddle.net/bz2s95ux/

I have a feeling it solves you problem without having to change in the engine of Alasql. Drawback: it adds a prototype to string. In the example its "addView" and "view" but in a multi tennant enviroment you would select something that would be more like a namespacing (like `addViewMyAwesomeFeatureNameSomethingLong2734687268762(...)`)

akhaneev:
Thank you! That's a great idea! It works with numbers as well. I think it will cover all cases. I'll try it in our product and will reply here then.

mathiasrw:
How did you go?

akhaneev:
Getting this error in our product: Cannot create property 'view' on string 'A'
We use webpack that seems to cause the issue. Had not much time to investigate the issue.

mathiasrw:
@akhaneev Are you doing `'A'.prototype.view = ...`?

akhaneev:
I do `'A'.setView(someView)` and it throws the error. If I override
`String.prototype.setView = function(view) {
  this.view = view;
  return this;
};`
in the browser console, then it works fine.

mathiasrw:
Hm. Can you clone the jsfiddle to show me?

akhaneev:
Not sure I can reproduce the issue in the jsfiddle, because the issue comes from webpack we use to build the app. But I found that `new String("A").setView("B")` works while `"A".setView("B")` throws the exception with this stack trace `at String.setView.String.setView` - double `String.setView`. And it looks fine in the compiled sources...

mathiasrw:
@akhaneev Ahh - perfect!

Ill close this PR then :)

akhaneev:
Hello @mathiasrw,
Sorry to bother you, but I discovered a problem: after I set some view to a string like `"a".setView("b")` it returns a String object. Then when it is compared in SQL, for example `... WHERE Column = "a"` it is always falsy, because it is compared strictly like `stringObject === "a"`. So this solution will work if any comparison in the alasql is not strict (`stringObject == "a"` returns true). Is there any setting for this in alasql?

mathiasrw:
@akhaneev I seeeee - hmmmmm. I have a feeling "123" will be equal to 123 then. Have to think about that.

akhaneev:
In my case, all columns in a table are typed as a string or a number, there is no mixed type. So comparing a string with a number should not be an issue.
For now, I see problems with these operators: `=`, `==`, `!=`, `!==`. If I replace them with triple =, the comparison works as expected from a user point of view. So I will use regex to replace those operators with triple = for the moment. But regex is not reliable, because I can replace = in a column name or in a string constant. So a default behavior in the alasql is preferred.
I remember there were ifs for those operators in the source code. If you think it is a good solution to make an option for a non-strict default comparison, I think I could create a PR.

PR #1219, opened 2020-08-28
String/Number object support

akhaneev:
Fix for cases when table data contains String and Number objects (not primitives). In this case typeof returns "object" that causes wrong calculation in MEDIAN and errors when AlaSQL creates copies of objects.
Not sure I found all cases with the issue.


mathiasrw:
👀

mathiasrw (approved):

mathiasrw:
Tank you!

PR #1225, opened 2020-09-24
JOIN USING compatibility with String and Number objects

akhaneev:
P.S. there is still incompatibility of operators like IN with String and Number objects. These operators use Array.indexOf() function which fails on String and Number objects. Maybe you have an idea of how it can be fixed?


mathiasrw:
Thank you!

mathiasrw:
> P.S. there is still incompatibility of operators like IN with String and Number objects. These operators use Array.indexOf() function which fails on String and Number objects. Maybe you have an idea of how it can be fixed?

Do you mean when you have `new String("abc")` as `name` and then do a `... where name in ('john', 'simon')`?

akhaneev:
Yes, exactly.

mathiasrw:
Released as part of version 0.6.4

PR #1252, opened 2020-11-10
Do not include null in COUNT or AVG fixes #1251

### LINKED ISSUE #1251: COUNT and AVG should not include null values.
`COUNT` and `AVG` should not include `null` values.
jsfiddle: http://jsfiddle.net/2obct5p9/5/

When using `COUNT` and `AVG` the `null` values are included, the code is only excluding `undefined` values. However normally one would expect in SQL that `null` cells are not included in a `COUNT` or `AVG`.

tastypackets:
Excludes `null` cells from `COUNT` and `AVG`.

It seemed a bit odd that this had to be updated in 3 areas to take effect to the `GROUP BY` as well, should prob be abstracted into a function or something simpler. I saw there was an issue talking about `v2` and I'm not familiar with the code base, so I thought it was better just to make the necessary updates instead of trying to refactor. 😅

### DIFF
--- a/src/423groupby.js
+++ b/src/423groupby.js
@@ -166,7 +166,15 @@ if(false) {
 						} else {
 							//						return "'"+colas+'\':(typeof '+colexp+' != "undefined")?1:0,';
 							//					} else {
-							return "'" + colas + "':(typeof " + colexp + ' != "undefined")?1:0,';
+							return (
+								"'" +
+								colas +
+								"':(typeof " +
+								colexp +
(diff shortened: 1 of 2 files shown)


mathiasrw:
Thank you! And you even included a test. So smooth.

mathiasrw:
Now released as part of v0.6.5

tastypackets:
> Now released as part of v0.6.5

Thank you for the quick turn around, it'll help me stay unblocked! 😁

PR #1279, opened 2021-01-08
Return empty results in group by when input is empty

### LINKED ISSUE #927: How to return empty array if no results?
Hi,
from my query that looks something like this:

```
SELECT column1, column2
FROM ?
GROUP BY column1, column2
ORDER BY column1 ASC 	
```
I'm getting an array with one object (where there should be no results found). That one object has those 2 columns as keys with values being `undefined`.

forestfang-stripe:
fixes #927 

We currently have a fallback check on empty `query.groups` so that we can populate aggregate function default value in a non-groupby SELECT. i.e. `SELECT COUNT(1) as c FROM ? WHERE 1 = 2` should return `[{c:0}]`.

However, `query.groups` can also be empty if there is no input data to group by. In this case, we would expect AlaSQL (and other SQL engines indeed) to not return anything instead of emitting an undefined group. Add an additional check to verify that `allgroups` are empty, that is, there were no GROUP BY clause.

### DIFF
--- a/src/38query.js
+++ b/src/38query.js
@@ -143,7 +143,7 @@ function queryfn3(query) {
 	//			console.log(query.havingfns);
 	if (query.groupfn) {
 		query.data = [];
-		if (0 === query.groups.length) {
+		if (query.groups.length === 0 && query.allgroups.length === 0) {
 			var g = {};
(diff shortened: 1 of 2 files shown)


mathiasrw:
Thank you!

PR #1419, opened 2022-01-27
fix: Let CONCAT_WS ignore null

alsundukov:
The concat_ws function is fixed according to the [documentation](https://www.w3schools.com/sql/func_mysql_concat_ws.asp):
"... An expression with a NULL value will be skipped"


mathiasrw:

PR #1428, opened 2022-02-16
fix: DATEADD does not take into account the value of the specified date

alsundukov:
The DATEADD function currently accepts uses fixed intervals, for example, for a MONTH interval it always uses 30 days. But months can be of different duration. The same with the QUARTER and YEAR.
Fixed it. Also fixed the DATEADD function test.

### DIFF
--- a/src/61date.js
+++ b/src/61date.js
@@ -130,8 +130,25 @@ alasql.stdfn.DATEDIFF = function (period, d1, d2) {
 };
 
 alasql.stdfn.DATEADD = function (period, interval, d) {
-	var nd = new Date(d).getTime() + interval * PERIODS[period.toLowerCase()];
-	return new Date(nd);
+	var nd = new Date(d);
+	var	period = period.toLowerCase();
+
+	switch (period) {
+		case "year":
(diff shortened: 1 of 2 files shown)


mathiasrw:
Looks like we got a daylight savings issue here - or a timezone issue. 

Not sure what the good solution is. Inputs welcome. 

image

image

alsundukov:
It looks like this is more of a test problem because toISOString() returns a string with no GMT offset. And datetime2 is initialized in the local time zone. Fixed this by adding GMT to the original date: `'2007-01-01 13:10:10.1111111 GMT'`.

mathiasrw:
Can you please take a look at https://github.com/AlaSQL/alasql/actions/runs/3135347924/jobs/5090938164#step:6:172 ?

looks like the quarter added is one hour off? tried to use utc to see if that would change anything. 

image

alsundukov:
It really looks like a problem with the time zone settings of the test environment. Fixed it for the nodeJS.

mathiasrw:
Released as part of v2.0.7

PR #1434, opened 2022-03-04
FIX #1433: IFNULL checks only undefined and null

### LINKED ISSUE #1433: IFNULL(a, b) returns b for falsy a values
Something is not working as expected:
- The function `IFNULL(a,b)` returns `b` even for all falsy JS values

An example of the bug can be found here: [http://jsfiddle.net/leobia/gkdo7zjw/](http://jsfiddle.net/leobia/gkdo7zjw/)

leobia:
Fixes IFNULL bug reported in #1433

Now IFNULL checks if the value is null or undefined. Other falsy values should pass the check

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -167,7 +167,7 @@ stdlib.IIF = function (a, b, c) {
 	}
 };
 stdlib.IFNULL = function (a, b) {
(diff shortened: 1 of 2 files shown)


mathiasrw:
Thank you!

PR #1449, opened 2022-03-30
Check for global `document` rather than `exports` with INTO HTML

brettz9:
This allows for one to set a `global.document = document` in Node via the likes of JSDom to query HTML one has parsed in Node (on the server side).

A better approach might be allowing the user to supply the document object as an option, but as it seems you don't really accept options in the constructor, I thought this might be as easy.

One other item this PR doesn't handle is that it is possible that `SELECT * FROM TXT('#one')` could be supported in Node too.

### DIFF
--- a/src/830into.js
+++ b/src/830into.js
@@ -71,7 +71,7 @@ alasql.into.SQL = function (filename, opts, data, columns, cb) {
 
 alasql.into.HTML = function (selector, opts, data, columns, cb) {
 	var res = 1;
-	if (typeof exports !== 'object') {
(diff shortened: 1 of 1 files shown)


brettz9:
btw, off topic, but an interesting idea for a user function would be one which could do `fetch` (in browser or in Node via `node-fetch`) to URLs and then grab tables out of the HTML, allowing for cross-site joins...

mathiasrw:
Awesome!

mathiasrw:

PR #1450, opened 2022-03-31
Eslint fixes

brettz9:
Some ESLint related fixes

- fix `globals` (is a top-level property, not a rule)
- add `overrides` for Mocha globals
- add lint script
- add eslint dependency
- add elsintignore for minified/built files
- ignore pnpm-lock.yaml file
- add editorconfig to enforce tabs in IDEs

Still have a ton of errors (including many that can be autofixed), but this might help orient for cleaning up.

Note that the test `Empty strings should be read as empty strings, not 0` in `/test/test621.js` is failing for me, mentioning a 404 (though I didn't change source code, and bumping the timeout didn't help).

I was also only able to install using pnpm, not yarn or npm. Getting this error:

> error An unexpected error occurred: "ENOENT: no such file or directory, copyfile '/Users/brett/Library/Caches/Yarn/v6/npm-escape-string-regexp-4.0.0-14ba83a5d373e3d311e5afca29cf5bfad965bf34-integrity/node_modules/escape-string-regexp/index.d.ts' -> '/Users/brett/alasql/node_modules/eslint/node_modules/escape-string-regexp/index.d.ts'".


mathiasrw:
Please run `npx yarn` to update the dependencies.

brettz9:
image

mathiasrw:
Looks like your yarn cache folder got out of sync. I suggest removing and installing yarn, or see if this works `yarn cache clean && yarn install`

brettz9:
The latter was sufficient in this case, thanks!

Issue #1426, opened 2022-02-15
AST to string causes double aliases


peterbe:
Steps to reproduce:
```js
import alasql from "alasql";

const sql = `
SELECT title, LENGTH(title) AS length FROM ?
`.trim()
const ast = alasql.parse(sql);
console.log(ast.toString())
// Will output:
// SELECT title, LENGTH(title) AS length AS length FROM $0 AS default
```
**Note the double `AS length`**

Demo here:
https://codesandbox.io/s/sleepy-poincare-5t0jd?file=/src/App.tsx

peterbe:
For my own rushing needs, here's how I temporarily hacked around it :)
```js
function fixDoubleAliases(sql: string): string {
  // Necessary because of https://github.com/agershun/alasql/issues/1426
  const dupes = new Set<string>();
  for (const match of sql.match(/\sAS \w+/g) || []) {
    if (dupes.has(match)) {
      sql = sql.replace(match, "");
    }
    dupes.add(match);
  }
  return sql;
}
```

mathiasrw:
Released as part of https://github.com/AlaSQL/alasql/releases/tag/v2.0.1

PR #1459, opened 2022-05-04
Fix: AST to string causes double aliases

Buroni:
Fixes https://github.com/AlaSQL/alasql/issues/1426

During `ast.toString()`, the function alias is being set in two places: At the `yy.FuncValue` level and the `yy.Select` level. This is causing double aliases to be generated for functions when stringifying the AST, e.g. `SELECT LENGTH(col) AS length AS length`. 

This problem would have existed on other tokens such as `yy.Literal`s too, but they seem to use an outdated(?) `this.alias` which always returns `undefined`. Updating the `yy.Literal` code to use `this.as` causes the problem to appear for literals too, e.g. `SELECT col AS c AS c`.

My solution is to just remove any alias handling-code from tokens which could represent columns, and let `yy.Select` handle those aliases as it currently does.

### DIFF
--- a/src/50expression.js
+++ b/src/50expression.js
@@ -159,12 +159,11 @@ yy.JavaScript.prototype.execute = function (databaseid, params, cb) {
 yy.Literal = function (params) {
 	return yy.extend(this, params);
 };
-yy.Literal.prototype.toString = function (dontas) {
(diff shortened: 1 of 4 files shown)


mathiasrw:
Thank you!

PR #1495, opened 2022-09-24
Add: TIMESTAMPDIFF from MYSQL

mathiasrw:
- Changes parser to allow syntax for MYSQL timestampdiff function
- Adds TIMESTAMPDIFF function to mysql.js, implementing using datediffThank you for the time you are putting into AlaSQL!


mathiasrw:

PR #1523, opened 2022-10-20
Fix newDate() and Now() reverted back the date format

alsundukov:
newDate() for Safari (et. al.)
Supported formats:
"2022.01.10 04:10",
"2022.01.10 04:10:11",
"2022.01.10 04:10:11.123",
"2022-01-10 04:10",
"2022-01-10 04:10:11",
"2022-01-10 04:10:11.123",
"2022.1.10",
"2022-1-1"
...

Fix for NOW() in Firefox: reverted back the date format: 2022-11-01 -> 2022.11.01

### DIFF
--- a/src/61date.js
+++ b/src/61date.js
@@ -51,9 +51,9 @@ stdfn.NOW = function () {
 	var d = new Date();
 	var s =
 		d.getFullYear() +
-		'-' +
+		'.' +
 		('0' + (d.getMonth() + 1)).substr(-2) +
-		'-' +
+		'.' +
 		('0' + d.getDate()).substr(-2);
 	s +=
 		' ' +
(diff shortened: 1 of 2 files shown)


mathiasrw:
Can you explain what is not good about the current implementation? What are we trying to solve? I am not a fan of supporting non-standard data formats. Considering adding a function or a flag to deal with this.

mathiasrw:
Yes. Breaking compability is a pain, but standadisation has an important role in making the world a better place. 

This is why we seek to make it clear to people that something is not compatible when we update the major per of the semver version of AlaSQL

image
image

Im happy to work with you on a solution if you can help me understand what is not working in safari

alsundukov:
OK, maybe you're right. Then adding an option to alasql.options for the date separator returned by the Now() function is a good solution.

PR #1548, opened 2022-11-28
Correction to utils.xlsnc to allow AAA and more

rcpacheco:
Currently the function utils.xlsnc does not calculate correctly columns beyond ZZ (position 701).
This change allos it to manage column names with 3 letters according to excel naming conventions.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v2.1.8

PR #1562, opened 2022-12-11
Fix sync XMLHttpRequest

jimmywarting:
I really do not like that you have logic for loading data in certain ways, like using fetch, xhr, fs etc
i think the developer should be responsible for loading them themself if they need to do it.

Just doing this to fix one test case. not that I'm dependent on this sync version... i really think this should be removed. it's better to just use fetch instead of xhr, would work in service workers too where xhr isn't available

using responseType = 'text' is not necessary, b/c that's the default anyways.
otherwise Chrome complains on `can't set responseType when using sync xmlhttprequest`

### DIFF
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -17,13 +17,13 @@ var utils = (alasql.utils = {});
   @function
   @param {string} s JavaScript string to be modified
   @return {string} Covered expression
- 
+
   @example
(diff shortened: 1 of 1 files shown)


mathiasrw:
Happy to work with you to make this more standard across all calls.

PR #1566, opened 2022-12-12
Drop gulp-dereserve

jimmywarting:
IE8 have been dead for a long time...

I also added this to match the version that are being tested in the github workflow. 
```json
"engines": {
  "node": ">=14"
},
```


jimmywarting:
> The resulting code can run in some strangely old versions of node.

I just thought it would be best to just simply add a version based on the current LTS. 
That way we could write things without having to worry about making any breaking changes.

I suppose that if you install `"alasql": "*"`, or run "npm install alasql" then npm will notice that you could install any version of alasql and if you have a older nodejs version then it would pick a previous release.

mathiasrw:
Good point

mathiasrw:
This is a breaking change, so will bundle a few changes before we release v3 now.

PR #1574, opened 2022-12-14
turn `Transaction` into a class

jimmywarting:
Also added eslint to dev dependency so that I could run `npm run lint`
knowing what was wrong showed me that some eslintrc config had to be changed to allow for things such as reseved class keyword didn't work and class fields

fyi, `commited` was changed to `committed` - hope that it don't break anything 

(got a [code spelling checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) extension in vscode)
having it on shows a LOT of blue squiggly lines over all files cuz your naming convention is more like: 

`transactionid` instead of: 
`transactionId`

it can read words that are CamelCase. It really helps me a lot of making such a silly mistake like this when english isn't your native spoken lanugage


jimmywarting:
you know what. this commit history was wack. starting over...

PR #1579, opened 2022-12-17
Fix idb

### LINKED ISSUE #877: indexedDB.webkitGetDatabaseNames() is deprecated and will be removed from Chrome
Just a heads-up that per https://www.chromestatus.com/features/5725741740195840 we're planning on removing the non-standard `indexedDB.webkitGetDatabaseNames()` API from Chrome in M60, around August 2017.

It looks like AlaSQL.js will handle this properly since it already functions on other browsers, but I wanted to check in and ensure that this deprecation timeline will not have any unforeseen impact.

### LINKED ISSUE #1577: create indexeddb table never runs callback
http://jsfiddle.net/Lcmkyfxe/

I'm debugging some browser test and one thing is timing out...
more specifically: 
```js
alasql('CREATE IndexedDB DATABASE ag154', [], () => {
  // this never runs...
})
```

Here is what i think happens:

https://github.com/AlaSQL/alasql/blob/f7984eab8862988ab815443fd0163d1a49b6f9d3/src/91indexeddb.js#L92-L97

1. `onupgradeneeded` is executed
2. the transaction gets aborted.
3. `request1.onabort` never gets called... (?)
4. an error is thrown but there is nothing that catches it... there are not any `request1.onerror`...
5. so no callback is executed.

I also notices that there are some throw exceptions... do they really work as they should?
shouldn't it be executing a callback?
I'm not sure what callback style you are using, it ain't like nodejs convention like `cb(err?, res)`
seems to be `cb(res?, err)`

I suppose this:
https://github.com/AlaSQL/alasql/blob/f7984eab8862988ab815443fd0163d1a49b6f9d3/src/91indexeddb.js#L98-L104
should be more like this?
```js
request1.onsuccess = function (event) { 
  event.target.result.close()
  cb(null, new Error(`IndexedDB: Cannot create new database "${ixdbid}" because it already exists`))
}
```

jimmywarting:
- `webkitGetDatabaseNames` got removed and replace with `indexeddb.showDatabases()`<br> so i replaced  `IDB.showDatabases` to use the new api.
- `IDB.deleteWhere` didn't pass in the 3rd alasql argument, so some test where failing there... that one argument got added.
- `CREATE IndexedDB DATABASE ag154 ` got fixed
- Some other small modernize in yy, statment and search. only refactor nothing really changed. search now use a class and a private class method instead (for the `doSearch`) don't know if this would be too breaking change or if it's even supported by our version range.
- tidy up html template test runner
- refactored two indexedb db test files to use alasql.promise for better readability.
- all other operation that relied on "if database exist" operations now uses my new  `databaseExists` function that checks if database exist and if not it test it by opening the named index db and aborts if it executes upgrade needed.
- some requests to the server requested files with two leading slashes `http://localhost:7387//test168.xlsx` the server now normalize it to one

closes #1577 
closes #877


jimmywarting:
Hmm, do you know why it failed? something about the port already being used?

jimmywarting:
ready to be reviewed / merged
not gona do any more commits to this now unless you request some changes

src/35search.js
```diff
@@ -8,1133 +8,1103 @@
...
+				if (cb) res = cb(res);
+			});
+			// if(cb) res = cb(res);
+			return res;
+		};
+		statement.query = {};
+		return statement;
+	}
+
+	#doSearch(databaseid, params, cb) {
```
mathiasrw, line 50: What is this `#` doing? Some ES6 tricks I dont know about?

src/35search.js
```diff
@@ -8,1133 +8,1103 @@
...
-							if (node.$out && node.$out.length > 0) {
-								node.$out.forEach(function (edgeid) {
-									var edge = objects[edgeid];
-									var stack2 = stack.concat(edge);
-									stack2.push(objects[edge.$out[0]]);
-									queue.push({
-										node: objects[edge.$out[0]],
-										stack: stack2,
-									});
+							return processSelector(selectors, sidx + 1, value);
```
mathiasrw, line 145: Looking so much forward to us massaging the code more to get out of if-else nested hell

test/test154.js
```diff
@@ -7,51 +7,40 @@ if (typeof exports === 'object') {
...
+
+			const res8 = await sql('SHOW TABLES FROM ag154');
+			assert(res8.length === 0);
+
+			const res9 = await sql('DETACH DATABASE ag154;DROP IndexedDB DATABASE ag154');
+			assert(res9[0] === 1);
+			assert(res9[1] === 1);
 		});
 	});
 }
```
mathiasrw, line 46: Such a good reminder about what async was put in place to solve...

src/35search.js
jimmywarting: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

PR #1581, opened 2022-12-19
Add content type response header

jimmywarting:
Web worker test are failing in chrome if it's not served with the correct content type.


mathiasrw:
Makes sense. Please rebase on develop (or merge) so we have no conflicts.

PR #1584, opened 2022-12-20
Better browser test

jimmywarting:
tough of making many commits/PRs, but there where so many files that i changed in the end with some variety.
To summarize what i did:

- i inlined and/or scoped some variable & functions inside of the describe function instead of having it in the global scope.<br> why? browsers run all the scripts with a `<script src>` So all the variable and functions are global and got in conflict of each other causing some test to fail when it run `Create database {test}` as it was using the same number in each and every file
- `deepStrictEqual` didn't exist, changed it to `deepEqual`...
- i tidy up some test to using async / await as i was skimming trough the failing browser test cases. 
- some test was purely NodeJS specific so i skipped them when not running in NodeJS

Other than that the test input/output haven't changed at all. 
There is currently now only one test that fails. And that one is: Test 408. Have something to do with dates, didn't really know how to solve it.

----

Little off topic below:

- it seems like CI/CD don't run browser test cases...? should try fixing that (after all test cases are working again)
- also i'm not a huge fan of the little [timeout snippet](https://github.com/AlaSQL/alasql/blob/1bb08d3638973b7ece183df2e1800f59ba9c3c69/test/browserTestRunner.tmlp.html#L29-L41) - would rather just want to remove it (and jQuery) all together. The test cases takes longer than 3000ms. i have kept increasing it or comment it durning my testing.
- Ultimately I would like to enable ESM mode by using: `<script type="module" src="...">` but i don't think we are there quite yet
- I think it would be a good idea to have separate folders for each environment specific test cases such as:
```
/test
|-- nodejs           <-- NodeJS specific 
|   |-- test001.js   <-- fs test
|-- browsers         <-- browser specific 
|   |-- test002.js   <-- indexeddb test
|-- test003.js       <-- universal
|-- test004.js       <-- universal
```


mathiasrw (approved):
Amazing work.

mathiasrw:
> it seems like CI/CD don't run browser test cases...? should try fixing that (after all test cases are working again)

Yes. Lets make a new issue with this. Not sure how best to do it. WOndering if we can use puppeteer https://pptr.dev/ to run the page but manage the test inside a node world. 

> also i'm not a huge fan of the little [timeout snippet](https://github.com/AlaSQL/alasql/blob/1bb08d3638973b7ece183df2e1800f59ba9c3c69/test/browserTestRunner.tmlp.html#L29-L41)

Find a way and lets change it. 

> Ultimately I would like to enable ESM mode

What do we gain from this?

> I think it would be a good idea to have separate folders for each environment specific test cases such as:

I agree. Also:  Lets do everything we can to avoid tests that are env. specific.

jimmywarting:
> Wondering if we can use puppeteer

I tough so too first, but then i remembered that it's just chrome/firefox specific. 
I have used Karama before that have different test runner that can start a variety of browsers installed locally so that it runs in IE, FF, Safari, Chrome, Edge, PhantomJS and so on...

I'm also asking myself: do we really need to test other browsers? is it good enough with just chrome? maybe it is.
i have been working on some WebTorrent submodules and they use [airtap](https://www.npmjs.com/package/airtap) to run in browsers

> Find a way and lets change it.

Could just simply remove the hole setTimeout all together? it's possible to click on the "failed" link in the upper right corner to filter stuff. it kind of do what the setTimeout is already doing.

> > Ultimately I would like to enable ESM mode
> What do we gain from this?

Then it would be more module like
And you wouldn't have to worry about polluting the `globalThis` (aka window)
then there are other things such as `import` statement that could in the end work the same way in both node and browser runtime. 
you wouldn't need to write any 
```js
if (typeof exports === 'object') { ... }
```
you can just as well have a [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) instead
and there is also the benefit of using top level await, but i suppose that we wouldn't use that at all even. but at least it's there.
Oh and then there is also `import.meta.url` that's helpful for loading files relative to xyz. so it would also help get rid of `__dirname` and kind of work the same way over both browser and node.

PR #1589, opened 2022-12-22
convert expressions to classes

jimmywarting:
Don't know how you have gotten away with calling `obj.constructor()` without using new.


mathiasrw:
Nice progress.

> Don't know how you have gotten away with calling obj.constructor() without using new.

By the magic of ES3

PR #1594, opened 2022-12-26
remove jquery

jimmywarting:


mathiasrw:
We should remove the files too

image

mathiasrw:
Actually - we have been long due to a general cleanup. Removed all files note used actively the last few years.

mathiasrw:
The merge conflict is there because I deleted the file

mathiasrw:
If we find someting that we are missing we just recreate those files.

PR #1600, opened 2022-12-28
Remove ie9 specific code

jimmywarting:
IE9 (and 10, 11) are long dead. nobody is using it anymore


mathiasrw:
How embarrassing we still had that in the code. Thank you!

PR #1603, opened 2022-12-31
Support the "total" aggregator to fix #1530

### LINKED ISSUE #1530: alasql.fn.total is not a function
_Spawn from https://github.com/AlaSQL/alasql-sqllogictest/blob/a10ff82e75edafae2ce96c9d02ab47cfc85d4507/results/output.md#3620-testevidenceslt_lang_aggfunctest_

    SELECT total(y) FROM t1
    
    alasql.fn.total is not a function

We would like to support TOTAL. Its almost the same as SUM but 

> This function is similar to the [sum()](https://database.guide/how-sqlite-sum-works/) function, [except in the way it handles NULL inputs](https://database.guide/sqlite-sum-vs-total-whats-the-difference/). When there are no non-NULL values, then sum() returns NULL (instead of 0.0 like the total() function returns).

https://database.guide/how-sqlite-total-works/

gangadhargo:
- Total function feature added.
- Refactored duplicate code in SUM, MAX, MIN

@mathiasrw  - would please check the **select.js** and **pivot.js** files?  I have added TOTAL where ever checks are related to SUM as both behave almost similarly. please let me know if it is fine or if any changes are required.


mathiasrw:
It looks good. Thank you so much for this @gangadhargo

mathiasrw:
Released as part of v2.5.0

PR #1607, opened 2023-01-08
Classify

jimmywarting:
Classify `yy.Select` and `yy.ExistValue`
Also moved `Select#compileWhereExists` and `Select#compileQueries` into select.js


mathiasrw:
Looks good to me.

PR #1618, opened 2023-01-17
fix: GROUP BY with CASE x IS NULL

alsundukov:
For case:

`SELECT COUNT(*) FROM ? GROUP BY CASE WHEN id IS NULL THEN alternativeId ELSE id END

>>> SyntaxError: Unexpected token ']'
`
Сause: in the file '423groupby.js' there are many places similar to col2.split('\t'), but in toJS() we use template literals with a lot of \t.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v2.5.2

Issue #1616, opened 2023-01-16
INSERT INTO with a SELECT * FROM and a subquery throws exception


hjkwon0609:
An `INSERT INTO` followed by a `SELECT` with a subquery throws an exception

```
>                                                                                                                                                                                                                                                                                                                             
> const alasql = require('alasql');                                                                                                                                                                                                                                                                                           
undefined                                                                                                                                                                                                                                                                                     
> alasql.exec('CREATE TABLE t (id INT, val TEXT)')                                                                                                                                                                                                                                                                            
1                                                                                                                                                                                                                                                                                                                             
> alasql.exec('CREATE TABLE t2 (id INT, val TEXT)')                                                                                                                                                                                                                                           
1                                                                                                                                                                                                                                                                                                                             
> alasql.exec('CREATE TABLE t3 (id INT, val TEXT)')                                                                                                                                                                                                                                                                           
1                                                                                                                                                                                                                                                                                                                             
> alasql.exec('INSERT INTO t2 values (1, "ABCDE")')                                                                                                                                                                                                                                                                           
1                                                                                                                                                                                                                                                                                                                             
> alasql.exec('INSERT INTO t3 values (1, "DSAFDSAFDS")')                                                                                                                                                                                                                                                                      
1                                                                                                                                                                        
>                                                                                                                                                     
> alasql.exec('INSERT INTO t3 values (2, "ABCDE")')                                                                                           
1                                                                                   
> alasql.exec('SELECT * FROM t2 WHERE val = (SELECT val FROM t3 WHERE id= 2)')                                                                                                                                                                                                                                               
[ { id: 1, val: 'ABCDE' } ]                                         
> alasql.exec('INSERT INTO t SELECT * FROM t2 WHERE val = (SELECT val FROM t3 WHERE id= 2)')                                                                                                                                                                                                                                  
Uncaught TypeError: Cannot read property '0' of undefined                                                                                                                                                                                                                                                                    
    at alasql.Query.eval [as wherefn] (eval at yy.Select.compileWhere (/home/mark/work/gopax/node_modules/alasql/dist/alasql.fs.js:8889:11), <anonymous>:3:74)                                                                                                                                                              
    at doJoin (/home/mark/work/gopax/node_modules/alasql/dist/alasql.fs.js:7449:13)                                                                                                                                                                                                                                         
    at doJoin (/home/mark/work/gopax/node_modules/alasql/dist/alasql.fs.js:7545:8)                                                                                                          
[...]

PR #1624, opened 2023-01-19
Fix (#1616): INSERT INTO with a SELECT * FROM and a subquery throws exception

ambujsahu81:
**What is the PR `for?**

- It is an attempt to fix the issue [#1616](https://github.com/AlaSQL/alasql/issues/1616)

**Description**  : 

 - It seems that the statement `alasql.exec('SELECT * FROM t2 WHERE val = (SELECT val FROM t3 WHERE id= 2)')` work fine without throwing any errors.  And the ast object has statement property which contain sql statement itself at index 0. this property contains `queries object`. Something like this :
```js
SELECT * FROM t2 WHERE val = SELECT val FROM t3 WHERE id = 2
statements: (1) [yy.Select]
      0: SELECT * FROM t2 WHERE val = SELECT val FROM t3 WHERE id = 2
            columns: (1) [Column]
            from:  (1) [Table]
            queries: (1) [yy.Select]    // <-   This property is present here
            where:  val = SELECT val FROM t3 WHERE id = 2
```
-  But for the statement `alasql.exec('INSERT INTO t SELECT * FROM t2 WHERE val = (SELECT val FROM t3 WHERE id= 2)')` `queries` property is present in `insert object` which also contain nested `select Object` where `queries` property is missing For example 
```js
INSERT INTO t SELECT * FROM t2 WHERE val = SELECT val FROM t3 WH
statements:(1) [yy.Insert]
	0: INSERT INTO t SELECT * FROM t2 WHERE val = SELECT val FROM t3 WH
	     into: t
	    queries: (1) [yy.Select]  // present here
	     select: SELECT * FROM t2 WHERE val = SELECT val FROM t3 WHERE id = 2
		columns:(1) [Column]
		from: (1) [Table]  // missing here
		where: val = SELECT val FROM t3 WHERE id = 2
```
- **To avoid the exception the code changes adds the missing `queries` property in `nested select` object**

### DIFF
--- a/src/70insert.js
+++ b/src/70insert.js
@@ -230,6 +230,9 @@ yy.Insert.prototype.compile = function (databaseid) {
 		// INSERT INTO table SELECT
 	} else if (this.select) {
 		this.select.modifier = 'RECORDSET';
+		if (this.queries) {
+			this.select.queries = this.queries;
+		}
 		var selectfn = this.select.compile(databaseid);
 		if (db.engineid && alasql.engines[db.engineid].intoTable) {
 			var statement = function (params, cb) {


mathiasrw:
Thank you so much!

mathiasrw:
Released as part of v2.5.3

PR #1629, opened 2023-02-02
fix #1415: UNION not returning correct result

### LINKED ISSUE #1415: UNION not returning correct result
AlaSQL is based on unpaid voluntary work. Thank you for taking the time to make it better. 

Question about how to ...
- Please use http://stackoverflow.com/questions/ask?tags=alasql

Something is not working as expected:
- Describe the problem  

    UNION does not work.

- Provide code that replicates the problem 
```
data1=[{a:'abc'},{a:'xyz'}]
data2=[{a:'123'},{a:'987'}]

alasql("select * from :a union select * from :b",{a:data1,b:data2})

```
--- FF 95.01 Linux
```
[
  {},
  {},
  {
    "a": "abc"
  },
  {
    "a": "xyz"
  }
]

```
alasql.version
"1.7.3-develop-0be167bcundefined"

----- Chrome 97.0 Linux
```

[
    {},
    {},
    {
        "a": "abc"
    },
    {
        "a": "xyz"
    }
]
```
alasql.version
'1.7.3-develop-0be167bcundefined'

---- Node
```

> data1=[{a:'abc'},{a:'xyz'}]
[ { a: 'abc' }, { a: 'xyz' } ]
> data2=[{a:'123'},{a:'987'}]
[ { a: '123' }, { a: '987' } ]
> 
> alasql("select * from :a union select * from :b",{a:data1,b:data2})
[ {}, {}, { a: 'abc' }, { a: 'xyz' } ]
> alasql.version
'1.7.3-develop-0be167bcundefined'

```

---- How does it work in sqlite3

```
sqlite> create table one(a text);
sqlite> create table two(a text);
sqlite> insert into a values('abc'),('def');
sqlite> insert into two values('xyz'),('123');
sqlite> select * from one union select * from two;
123
abc
def
xyz


```

- We suggest to spawn a jsfiddle from http://jsfiddle.net/98qL7m0c/

----

Maybe I'm using alasql incorrectly? I tried looking at some documentation in the wiki but the section on union is quite spartan.

Anyway, thanks a lot!

ambujsahu81:
**What is the PR `for?**
- This is an attempt to fix the issue [#1415](https://github.com/AlaSQL/alasql/issues/1415).
- It seems that empty objects where due to `query.columns.length = 0`. That is why `res` and `res2` had empty objects like this `{}`. May be the solution should make sure that `query.columns.length !=0`.  
- The solution in this pr might be a bad approach to resolve the issue. But if there are any feedback or issues related to my current solution than i could try to work on different approach too.🙂

**Examples :**
```js
var data1=[{a:'abc'},{a:'xyz'}];
var data2=[{a:'123'},{a:'987'}];
var res = alasql("select * from :a union select * from :b",{a:data1,b:data2});

console.log(res); // [ { a: '123' }, { a: '987' }, { a: 'abc' }, { a: 'xyz' } ]

var res2 = alasql("select * from @[{x:true}, {x:3}] UNION select * from @[{x:false}, {x:9}]");

console.log(res2); // [ { x: false }, { x: 9 }, { x: true }, { x: 3 } ]

alasql("CREATE TABLE one (a string)");
alasql("CREATE TABLE Two (a string)");
alasql("INSERT INTO one VALUES ('abc'),('xyz')");
alasql("INSERT INTO Two VALUES ('123'),('987')");

var res3 = alasql("SELECT * FROM one union SELECT * FROM Two");

console.log(res3) // [ { a: 'abc' }, { a: 'xyz' }, { a: '123' }, { a: '987' } ]
```

### DIFF
--- a/src/38query.js
+++ b/src/38query.js
@@ -242,10 +242,17 @@ function queryfn3(query) {
 			ilen = nd.data.length;
 			for (var i = 0; i < ilen; i++) {
 				r = {};
-				jlen = Math.min(query.columns.length, nd.columns.length);
-				for (var j = 0; j < jlen; j++) {
-					r[query.columns[j].columnid] = nd.data[i][nd.columns[j].columnid];
-				}
+				if (query.columns.length) {
+					jlen = Math.min(query.columns.length, nd.columns.length);
+					for (var j = 0; j < jlen; j++) {
+						r[query.columns[j].columnid] = nd.data[i][nd.columns[j].columnid];
+					}
+				} else {
+					jlen = nd.columns.length;
+					for (var j = 0; j < jlen; j++) {
+						r[nd.columns[j].columnid] = nd.data[i][nd.columns[j].columnid];
+					}
+				}			
 				ud.push(r);
 			}
 		}
--- a/src/38query.js
+++ b/src/38query.js
@@ -252,7 +252,7 @@ function queryfn3(query) {
 					for (var j = 0; j < jlen; j++) {
 						r[nd.columns[j].columnid] = nd.data[i][nd.columns[j].columnid];
 					}
-				}			
+				}
 				ud.push(r);
 			}
 		}
--- a/test/test1415.js
+++ b/test/test1415.js
@@ -1,0 +1,27 @@
+if (typeof exports === 'object') {
+	var assert = require('assert');
+	var alasql = require('..');
+} else {
+	__dirname = '.';
+}
+
+describe('Test 1415 - UNION Expression with empty query columns bug', function () {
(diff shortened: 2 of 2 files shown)


mathiasrw:
Interesting. Thank you so much for taking the time do dive into this. -

Any chance you can add a test file so we know if we break it again one day?

ambujsahu81:
> Interesting. Thank you so much for taking the time do dive into this. -
> 
> Any chance you can add a test file so we know if we break it again one day?

@mathiasrw , yes I will try to add a test file for this fix today 🙂

mathiasrw:
Released as part of v3.1

Issue #1513, opened 2022-10-05
ORDER BY and dates issue


fabads:
Hello, 
I encounter an issue trying to ordering my data containing date columns.

Here an example of code (Apps script):

```
function myTest() {
  var data =[{id: 200, fromDate: new Date("2022-11-29"), toDate: new Date("2022-11-29")}, 
    {id:  100, fromDate: new Date("2022-11-30"), toDate: new Date("2022-11-30")},
    {id:  101, fromDate: new Date("2022-11-30"), toDate: new Date("2022-11-30")}, 
    {id:  102, fromDate: new Date("2022-11-30"), toDate: new Date("2022-11-30")}, 
    ];

  var db = new alasql.Database("MyDB");
  db.exec('CREATE TABLE mytable (id INT, fromDate Date, toDate Date)');
  db.tables.mytable.data = data;

  console.log(db.tables.mytable.data);
  var res = db.exec(`SELECT id FROM mytable ORDER BY fromDate, id ASC`,);

  console.log(res);
}
```

The result I get:

`[ {id: 200}, {id: 102}, {id: 101}, {id: 100} ]`

But the result I expect is:

`[ {id: 200}, {id: 100}, {id: 101}, {id: 102} ]`

Any idea of what's happening ?

fabads:
With the JSFiddle:
[http://jsfiddle.net/crf3nvj6/](url)

fabads:
Yes you are right, deeply linked to the Javascript date format. I tried to use string date (YYYYMMDD) and the double ordering is working fine. In my case, I do not have the choice, my data source contains javascript dates already.

Buroni:
I've spent some time looking into this issue. It seems like there's several factors involved.

This seems to be partially fixed on the `develop` branch for `2.1.3`. If I recreate @fabads' example in a local test on `develop`, it reproduces the error. However if I add `fromDate` to the select query columns then it works as we'd expect.

```sql
# This doesn't work on `develop`
SELECT id FROM mytable ORDER BY fromDate, id ASC

# This does work
SELECT id, fromDate FROM mytable ORDER BY fromDate, id ASC
```

[...]

PR #1649, opened 2023-03-10
Fix `order by` dates in queries with incomplete type information

Buroni:
Fixes https://github.com/AlaSQL/alasql/issues/1513

## Proposed changes

- Set `alasql.options.valueof = true` by default. This was originally `false` for [backwards compatibility reasons](https://github.com/AlaSQL/alasql/issues/1513#issuecomment-1461128150), but causes problems when sorting dates in anonymous queries and queries where the `SELECT` result doesn't include the date column used in the `ORDER BY` clause (E.g. `SELECT i FROM tbl ORDER BY date, i`).
- Small refactor of order by logic to always use `valueof` in `ParamValue`s and `Column`s when `options.valueof = true;`. It seems like this was the original intention but the `options.valueof` check was being done in the wrong place.

## Notes

- It looks like `yarn format` has also touched some files not related to this work.
- There's some duplication in the orderby code that could be refactored into something nicer I think. Although I've followed the existing pattern for now to make the solution clearer.


mathiasrw:
Thank you!

mathiasrw:
We need to bump the major version so will bundle this change with a few others we have been talking about.

mathiasrw:
This has not been forgotten...

Issue #1641, opened 2023-02-18
In INDEXEDDB, alasql.promise fails to create multiple tables with sql in array, but works with sql as string


arnemorken:
When creating multiple tables with the CREATE statements in an array of strings, only the first (nonexisting) table is created.
When concatenating all the CREATE statements in the array into one string, all the tables are created.

In the first case, the code below will fail with 
```
"alasql.min.js:48 Uncaught DOMException: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.
    at u.onsuccess (http://localhost/Prosjekter/testserver/wp-content/plugins/anyvista/view/data/alasql.min.js:48:108479)
u.onsuccess @ alasql.min.js:48"
```
If using LOCALSTORAGE instead of INDEXEDDB, both methods will work.

Code to reproduce (using _AlaSQL v3.0.0 build: develop-4287740e_):
```
<!doctype html>
<html lang="en">
<head>
<meta charset="utf-8" />
<script src="./alasql.min.js"></script>
<script>
// Create and connect database
alasql.promise(
  "CREATE INDEXEDDB DATABASE IF NOT EXISTS alatest;"+
  "ATTACH INDEXEDDB DATABASE alatest;"+
  "USE alatest"
)
.then( function(res) {
  // Create tables
  return alasql.promise(
    /*
    // This works:
    "CREATE TABLE IF NOT EXISTS mytable1 ( myid STRING, myname STRING );\
     CREATE TABLE IF NOT EXISTS mytable2 ( myid STRING, myname STRING );",
    */
    // This should work too, but doesnt:
    [
    "CREATE TABLE IF NOT EXISTS mytable1 ( myid STRING, myname STRING )",
    "CREATE TABLE IF NOT EXISTS mytable2 ( myid STRING, myname STRING )",
    ]
  );
})
.then( function(res) {
  // Fill tables
  return alasql.promise(
    [
    "INSERT INTO mytable1 (myid,myname) VALUES ( '1', 'Mr. One' )",
    "INSERT INTO mytable2 (myid,myname) VALUES ( '2', 'Mr. Two' )", // Fails if using second CREATE method above
    ]
  );
})
.then( function(res) {
  return alasql.promise("SELECT * from mytable1");
})
.then( function(data1) {
  console.log("data1:");
  console.log(data1);
});
</script>
```

ParardhaSarmah:
Hi. Can I work on this issue?

dcode333:
[...]

PR #1721, opened 2023-06-11
Added fix for IndexDB version not incrementing due to parallal execution of sql statements

Julias0:
Fix for #1641 

Inside 91indexeddb.js, 
```javascript
const request = indexedDB.open(ixdbid, found.version + 1);
```

Since PromiseAll triggers all of the statements at the same time, it causes version to get incremented only once even if multiple DDL statements are present. Have modified the behaviour to execute the queries sequentially which will prevent race conditions such as these from coming up


mathiasrw:
This is such a good point. Thank you!

Any chance you can move all the functionality into the promiseAll function? Its doing what promiseAll should have been doing all the time.

test/browserTestRunner.js
```diff
@@ -37,7 +37,7 @@
 		return results;
 	}
 
-	const testFiles = walkFiles(__dirname, /test\d{3}\.js$/, null, true, true);
+	const testFiles = walkFiles(__dirname, /test\d{3,4}\.js$/, null, true, true);
```
Julias0, line 40: The tests with 4-digit numbers were not being run inside the browser

Julias0:
> This is such a good point. Thank you!
> 
> Any chance you can think about how we add a few tests for this?

Have added a new test to check this scenario @mathiasrw Lemme know if it works.

mathiasrw:
I really like how precise you nailed this problem.

Issue #1645, opened 2023-02-22
Bug when a variable is assigned from a query


pabloBelt:
Hi, 

There is a bug when a variable takes a value from a query.

`DECLARE` @X NUMBER = (SELECT MAX(population) FROM cities) `

http://jsfiddle.net/wa5hz8nd/1/

It can also be reproduced with the latest 3.0.0 version:

![image](https://user-images.githubusercontent.com/125101677/220732094-35f4d4f4-28f2-4215-9931-0aed28f40284.png)

Bests,
Pablo

joan8ruth:
Hello @pabloBelt ,
I see that the declare could be split into two steps:

declare @V NUMBER; 
SET @V = (Select MAX(X) from T);

and is working fine this way (maybe it works only this way? because I believe this is what I usually do when writing queries), but I would like to know if the single line declaration you have mentioned was working in the previous version.

Regards,
Joanna

abhishek-kumar-91:
I want to work on this issue, please assign me.

mathiasrw:
Fix Released as part of v4.0.6

PR #1725, opened 2023-06-16
Added a fix for the variable assignment not working properly

Julias0:
Added a fix for issue #1645 

The reference to `this` was getting modified to the window object which led to the statement not being able to resolve properly. Changing that fixed the issue.

Have also replaced the `.map` with a `.forEach` since the map is being used like a forEach

### DIFF
--- a/src/77declare.js
+++ b/src/77declare.js
@@ -36,8 +36,9 @@ yy.Declare.prototype.toString = function () {
 
 yy.Declare.prototype.execute = function (databaseid, params, cb) {
 	var res = 1;
-	if (this.declares && this.declares.length > 0) {
-		this.declares.map(function (declare) {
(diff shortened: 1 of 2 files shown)


src/77declare.js
```diff
@@ -36,8 +36,9 @@ yy.Declare.prototype.toString = function () {
 
 yy.Declare.prototype.execute = function (databaseid, params, cb) {
 	var res = 1;
-	if (this.declares && this.declares.length > 0) {
-		this.declares.map(function (declare) {
+	var that = this; // without this assigned to a variable, inside the forEach, the reference to `this` is lost. It is needed for the Function statement for binding
```
mathiasrw, line 39: Nice spotted!

mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.0.6

Issue #1400, opened 2021-11-19
Parsing CSV seems to force all value in column to have same format?


niwasmala:
# Describe the problem
I have a table like this on [google sheet](https://docs.google.com/spreadsheets/d/1je7bZV-YKek2S27-6mjL09OQ60Aefj8i6sUp5_a8Giw/edit?usp=sharing). Please note the account number some number have 0 as first value. I already set as plain text in google sheet.

![image](https://user-images.githubusercontent.com/476869/142577083-82c62824-fc89-4f7e-b47a-43c553bc5ad2.png)

When I access the sheet as csv using [this url](https://docs.google.com/spreadsheets/d/1je7bZV-YKek2S27-6mjL09OQ60Aefj8i6sUp5_a8Giw/export?format=csv&gid=1024258862), I got the zero as expected. 

![image](https://user-images.githubusercontent.com/476869/142577363-c5d3813a-46d6-432f-8c5f-c392bd8c2aa0.png)

But when I run the query, seems like AlaSQL force to return the value as number so the zero is missing.

```
async function loadPayment() {
  const url = 'https://docs.google.com/spreadsheets/d/1je7bZV-YKek2S27-6mjL09OQ60Aefj8i6sUp5_a8Giw/export?format=csv&gid=1024258862';
  const res = await alasql.promise('SELECT * FROM CSV(?, {headers:true})', [url]);

  console.log('payment', res);

  return res;
}

(async () => {
	await loadPayment();
})();
```

Expected: The account number display as it shows on csv (plain text)

Actual: The zero value is missing. Seems like the account number are forced as number?

![image](https://user-images.githubusercontent.com/476869/142577608-60837306-6263-4422-9265-2b2d2fd2a18b.png)

# Code replicates the problem
https://jsfiddle.net/evasdoxf/

mathiasrw:
As I recall this would not happen if the fields had a `"..."` wrapper. 

Hmm. I guess it would be good to disable to auto discovery feature of format.

Julias0:
I can take a stab at adding this feature

mathiasrw:
"The account number display as it shows on csv (plain text)"

The value is not encapsulated in `"..."` and should really be considered a number. 

I agree that if this is loaded into a table that says "String" as type, then this is the correct value. -

What about an option to retreive all data as strings, so you can do your conversion after?

mathiasrw:
[...]

PR #1729, opened 2023-06-19
Issue 1400

Julias0:
This contains feature for a new option for csv imports called `raw`. Passing that should make sure that the imported values are all strings solving the problem for #1400 

There is also a bug fix that if CSVs were imported without headers, the numerical values inside the csv would not be parsed into numbers.


src/84from.js
```diff
@@ -263,7 +264,7 @@ alasql.from.CSV = function (contents, opts, cb, idx, query) {
 						hs.forEach(function (h, idx) {
 							r[h] = a[idx];
 							// Please avoid === here
-							if (typeof r[h] !== 'undefined' && r[h].length !== 0 && r[h].trim() == +r[h]) {
+							if (!opt.raw && typeof r[h] !== 'undefined' && r[h].length !== 0 && r[h].trim() == +r[h]) {
```
Julias0, line 267: If the option raw is passed, we ignore parsing to number

src/84from.js
```diff
@@ -181,6 +181,7 @@ alasql.from.CSV = function (contents, opts, cb, idx, query) {
 		separator: ',',
 		quote: '"',
 		headers: true,
+		raw: false
```
Julias0, line 184: For backward compatibility

mathiasrw:
Released as part of v4.1.0

Issue #1547, opened 2022-11-28
SELECT RECORDSET * from csv returning empty columns if there's no matched rows


jchnxu:
<img width="1397" alt="image" src="https://user-images.githubusercontent.com/4531670/204205873-caf44a20-f50d-41ce-9c2c-efeb3bafeb6e.png">

Reproducing link: https://replit.com/@jchnxu/alasql-recordset-bug#index.js

aleeekhaan:
Hi can I work on this??

mathiasrw:
Absolutely

mathiasrw:
Any luck @aleeekhaan ?

aleeekhaan:
Working on it

mathiasrw:
Let me know if you need help @aleeekhaan

aleeekhaan:
Hi @mathiasrw,

I have resolved the issue. Should I open a PR, so that the changes can be reviewed?

mathiasrw:
@aleeekhaan yes please!

aleeekhaan:
A PR has been opened, #1733, please review code changes and let me know if any thing else is required.

mathiasrw:
Fix released in v4.11.0

PR #1733, opened 2023-06-28
fix RECORDSET returning empty columns if there's no matched rows issu…

aleeekhaan:
…e #1547

### DIFF
--- a/src/40select.js
+++ b/src/40select.js
@@ -468,6 +468,11 @@ function modify(query, res) {
 		} else {
 			// Cannot recognize columns
 			columns = [];
+			if (query && query.sources) {
+				query.sources.forEach((source) => {
+					columns = columns.concat(source.columns)
+				})
+			}
 		}
 	}
 


mathiasrw:
Lovely!

Any chance you can add a test file for this so we dont accidentally introduce the problem again when we rewrite things in a few years?

aleeekhaan:
Sure. I'll add a test case for this as well.

aleeekhaan:
Hi @mathiasrw, I have edited the test case you created to use a custom function for checking that the columns exist instead on using deepEqual. Please review and let me know if any changes are required.

mathiasrw:
I see the idea - just to make sure the columns are there - but we cant suddenly provide a different response format. 

Please refert to deep equal and see if you can get it to return columnId without 

```
  "dbenum": [undefined]
  "dbprecision": [undefined]
  "dbsize": [undefined]
  "dbtypeid": "INT"
```

aleeekhaan:
@mathiasrw The reason deepEqual fails because in the test case we only give the columnid while the actual result also has some other fields 

`  "dbenum": [undefined]
  "dbprecision": [undefined]
  "dbsize": [undefined]
  "dbtypeid": "INT"
  `

Can we remove these fields by writing a function that modifies the response for the recordset. Will this approach be appropriate?

mathiasrw:
Yep. That is what I was working towards with https://github.com/AlaSQL/alasql/pull/1733/commits/4feca50112b169df6eb7d8e262867dec90e654d8#diff-daa185eed55b4159786a570955a6a87c12074d6ed316e7c14bd1ef711aabf0d8R474

harisiqbal12:
@mathiasrw What's the update on this?

mathiasrw:
@harisiqbal12 Hmmm. Good question. Let me look into this over the weekend.

mathiasrw:
If anyone can pick this up that would be amazing.

mathiasrw:
Lifting this in, to get help solving the last little things.

Issue #1409, opened 2021-12-14
Triggers not working with IndexedDB


jiron12:
On an IndexDB table triggers are not firing.

Code example:

```
    alasql.fn.onInsert = function (r) {
        console.log('this never happens!');
    };

    await alasql.promise(`CREATE INDEXEDDB DATABASE IF NOT EXISTS test_db`);
    await alasql.promise(`ATTACH INDEXEDDB DATABASE test_db; USE test_db;`);

    await alasql.promise(`DROP TABLE IF EXISTS asset7`);
    await alasql.promise(`CREATE TABLE asset7([id] varchar(36) NOT NULL,  [name] varchar(45) NOT NULL, PRIMARY KEY ([id]) );`)

    var data = [
        { id: crypto.randomUUID(), name: "test1", amount: 7 },
        { id: crypto.randomUUID(), name: "test2", amount: 8 },
        { id: crypto.randomUUID(), name: "test3", amount: 9 }
    ];
    await alasql.promise(`INSERT INTO asset7 SELECT * FROM ?`, [data])

    await alasql.promise(`CREATE TRIGGER mytrigger after INSERT ON asset7 onInsert`)

    var data2 = [
        { id: crypto.randomUUID(), name: "test17", amount: 17 }
    ];
    await alasql.promise(`INSERT INTO asset7 SELECT * FROM ?`, [data2]);
```

Julias0:
@mathiasrw can you assign this to me

mathiasrw:
Fix released as part of v4.1.2

PR #1735, opened 2023-06-29
Added support for post insert triggers for indexdb

Julias0:
Fix for issue reported inside #1409 

Added support for post insert hook for indexedDb Databases.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.1.2

Issue #1556, opened 2022-12-10
query from indexeddb that has no `keyPath` (columns) returns weird data


jimmywarting:
tried executing this:
```js
await alasql.promise(`SELECT * FROM [table_name]`)
```
...on a existing indexeddb table that have no keyPaths. the result looked kind of like:

```js
{0: '1', 1: '0', 2: '4', 3: '4', 4: '4', 5: '1', 6: '1', 7: '0', 8: '1' ... }
```

just looking at the table in chrome devtool shows that it is just a key value pair without an object
![image](https://user-images.githubusercontent.com/1148376/206875233-dffc138c-63b7-4a6d-9ce1-c5b66bf1d41a.png)

jimmywarting:
found out that i could run: 
```js
await alasql.promise(`SELECT value FROM [table_name]`)
```
and then it would return the value correctly in a nice looking array.
but how can i get the keys?

jimmywarting:
I see that you are pushing the value to an array, but you never do anything else with the key, it's totally discarded
https://github.com/AlaSQL/alasql/blob/6e108289e8f23da1e10eca81100db0ff269550b1/src/91indexeddb.js#L436

didn't find any other place where you return the key

vishal6557:
@mathiasrw Is this available to work on? I know it's pretty old, but just double-checking. This would be my first contribution, and I am ready to dive in.

mathiasrw:
Fix released at part of v4.1.3

PR #1745, opened 2023-07-13
fix(IndexedDB): Added fix when there is no keyPaths(columns) in indexDB

vishal6557:
I have added a fix for the issue: #1556 .
If the keyPath is not present in the value, then I'm creating a object with rowKey as key of the object and returning as ` {shell_id_key: 'random_value2'}` where `shell_id_key` is the rowKey. 

Screenshot 2023-07-13 at 6 38 01 PM


mathiasrw:
Thank you!

mathiasrw:
Released at part of v4.1.3

PR #1767, opened 2023-07-30
Devops

### LINKED ISSUE #1766: Cut size by not polyfilling promoses
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#browser_compatibility promises are provided by all env. we are supporting. 

Lets cut out the polyfil

mathiasrw:
Clean up code and dependencies


mathiasrw (commented):
A bit messy - but it works.

PR #1771, opened 2023-08-12
feat: add from collection to AlaSQL interface type

sytone:
Add from collection to AlaSQL interface type along with the two from interfaces so modification of from in TypeScript is checked and allowed.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.1.4

PR #1778, opened 2023-08-18
alasql distinct (lowercase) with aggregation functions

### LINKED ISSUE #1776: alasql distinct (lowercase) with group_concat
Fiddle Example:  https://jsfiddle.net/udynw6k9/

We see alasql('SELECT GROUP_CONCAT(**distinct** answeredDate) from ? does not work, where as SELECT GROUP_CONCAT(DISTINCT answeredDate) from ? does work. 

The thing is that it behaves like it no oped on `distinct`  in above query.

But queries like `SELECT distinct answeredDate from ?` or `SELECT DISTINCT answeredDate from ?` work fine.

The GROUP_CONCAT is just for example, It probably happens with any of other aggregations with distinct as I had same issue with a custom aggr.
.

rrameshkumar76:
Fixes https://github.com/AlaSQL/alasql/issues/1776

### DIFF
--- a/src/alasqlparser.jison
+++ b/src/alasqlparser.jison
@@ -1376,7 +1376,7 @@ FuncValue
 					$$ = new yy.FuncValue({funcid: funcid, args: exprlist});
 			} else if(alasql.aggr[$1]) {
 		    	$$ = new yy.AggrValue({aggregatorid: 'REDUCE',
-                      funcid: funcid, expression: exprlist.pop(),distinct:(($3+'').toUpperCase()) });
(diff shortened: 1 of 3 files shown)


mathiasrw (commented):
Turns out the alasqlparser.js is autogenerated - so you need to change this in the alasqlparser.jison file 

https://github.com/AlaSQL/alasql/blob/develop/src/alasqlparser.jison#L1379

Use `yarn jison && yarn test` to confirm it is working.

rrameshkumar76:
I will fix that and update here. Also will rollback the change to alasqlparser.js

rrameshkumar76:
Sorry due to some issues, I opened a new PR https://github.com/AlaSQL/alasql/pull/1781

PR #1782, opened 2023-08-22
issue fixed #1740

### LINKED ISSUE #1740: Parse() then AST.toString() doesn't restore square brackets on column name
If a column name requires square brackets, e.g. due to a space or a period in the column name, toString() on the AST loses the brackets resulting in incorrect SQL.

For example
```
const ast = alasql.parse('SELECT [Foo Bar] FROM tbl');
console.log(ast.toString());
```
gives `SELECT Foo Bar FROM tbl` instead of `SELECT [Foo Bar] FROM tbl` 

Similarly:
```
const ast = alasql.parse('SELECT [Foo.Bar] FROM tbl');
console.log(ast.toString());
```
gives `SELECT Foo.Bar FROM tbl` instead of `SELECT [Foo.Bar] FROM tbl`

akashyadav24:
fixed #1740


mathiasrw:
Awesome! 

Tests are failing. Is `yarn test` working locally for you?

Also: any chance you can add a test file named test/test1740.js that asserts the issue being fixed so we know if we break this in the future?

mathiasrw:
@akashyadav24 how can I help you get the last little bit aligned?

mathiasrw:
Lifting this to better help with the last little bits.

PR #1794, opened 2023-09-11
Let LIKE work correctly with line breaks in values to fix #1734

### LINKED ISSUE #1734: LIKE will not work correctly when there is a line break
Run this code on [RunKit](https://npm.runkit.com/alasql)
Example Code:
```JavaScript
var alasql = require("alasql")
var data = [ {a: 'Lorem \n\r ipsum'} ];
var res = alasql('SELECT a FROM ? WHERE a LIKE "Lorem%"',[data]);
```
Expect Output:
> [{a: "Lorem \n\r ipsum"}]

Actual Output:
> []

When I change code to this, It output correct result:
```JavaScript
var alasql = require("alasql")
var data = [ {a: 'Lorem \n\r ipsum'} ];
var res = alasql('SELECT a FROM ? WHERE REPLACE(REPLACE(a, CHAR(13), ""), CHAR(10), "") LIKE "Lorem%"',[data]);
```

rameshrajagopalanbayer:
Fix #1734 - LIKE will not work correctly when there is a line break.

### DIFF
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -1197,7 +1197,7 @@ var like = (utils.like = function (pattern, value, escape) {
 		} else if (c === '[' || c === ']') {
 			s += c;
 		} else if (c === '%') {
-			s += '.*';
+			s += '[\\s\\S]*';
 		} else if (c === '_') {
 			s += '.';
 		} else if ('/.*+?|(){}'.indexOf(c) > -1) {
@@ -1211,7 +1211,7 @@ var like = (utils.like = function (pattern, value, escape) {
 	s += '$';
 	//    if(value == undefined) return false;
 	//console.log(s,value,(value||'').search(RegExp(s))>-1);
-	return ('' + (value || '')).toUpperCase().search(RegExp(s.toUpperCase())) > -1;
+	return ('' + (value || '')).search(RegExp(s, 'i')) > -1;
 });
 
 utils.glob = function (value, pattern) {
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -1211,7 +1211,7 @@ var like = (utils.like = function (pattern, value, escape) {
 	s += '$';
 	//    if(value == undefined) return false;
 	//console.log(s,value,(value||'').search(RegExp(s))>-1);
-	return ('' + (value || '')).search(RegExp(s, 'i')) > -1;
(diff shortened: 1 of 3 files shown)


test/test1734.js
```diff
@@ -0,0 +1,23 @@
...
+var test = '1666';
+describe('Test'  + test + 'Newline characters in like', function () {
+	it('1. LIKE', function (done) {
+		var data = [
+			{a: 'one', b: 'first'},
+			{a: 'two', b: 'second\n\ritem'},
+			{a: 'THREE', b: 'THIRD'},
+		];
+
+		var res = alasql('SELECT b FROM ? WHERE a LIKE "T%"', [data]);
```
mathiasrw: Dont you need to do the like operation on the b values - to make sure it correctly matches the value that includes a new line?

src/15utility.js
```diff
@@ -1197,7 +1197,7 @@ var like = (utils.like = function (pattern, value, escape) {
 		} else if (c === '[' || c === ']') {
 			s += c;
 		} else if (c === '%') {
-			s += '.*';
+			s += '[\\s\\S]*';
```
mathiasrw, line 1200: Ohhh, super nice way to solve this.

src/15utility.js
```diff
@@ -1211,7 +1211,7 @@ var like = (utils.like = function (pattern, value, escape) {
 	s += '$';
 	//    if(value == undefined) return false;
 	//console.log(s,value,(value||'').search(RegExp(s))>-1);
-	return ('' + (value || '')).toUpperCase().search(RegExp(s.toUpperCase())) > -1;
+	return ('' + (value || '')).search(RegExp(s, 'i')) > -1;
```
mathiasrw: Nice clean up. 

Looks to me like we have a bug if the value = 0 

Please use the nullish operator like this `(value ?? '').search`

mathiasrw (approved):

src/15utility.js
rameshrajagopalanbayer: sure will fix that later today and update here.

rameshrajagopalanbayer: I have changed that and added a test

mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.1.8

Issue #1789, opened 2023-09-07
JOINs only return the second subquery


gygabyte017:
Consider this simple query:

```
alasql(`	
	SELECT 
		a.outcome, 
		a.n, a.n/b.n as p 
	FROM (
		SELECT 	
			result as outcome, 
			COUNT(*) AS n 
		FROM ? 
		GROUP BY result
	) a 
	CROSS JOIN (
		SELECT COUNT(*) as n FROM ?
	) b
`, [x, x])
```

It returns this kind of object: `[{n: 123}]`, so it seems to be returing just the result of the subquery `SELECT COUNT(*) as n FROM ?` instead of the whole query, how is that possible?

It consistently happens as well with other types of JOINs

mathiasrw:
> this simple query

Welll, lots of things going on there in my opinion :)

> It returns this kind of object: [{n: 123}]

Clearly a bug. Thank you for taking the time to let the world know.

adarshjhaa100:
Hi @mathiasrw I could diagnose this issue and if I find something will raise a PR if nobody's working on this already.

mathiasrw:
Thank you!

adarshjhaa100:
@mathiasrw I've identified the issue. Whenever we have a subquery inside the join, the source.datafn is equal to the following function:
![image](https://github.com/AlaSQL/alasql/assets/31096082/26d3cafd-5d8e-4b94-9302-e1aeb9331d92)

The problem with just using the `subquery` function here is that even though it returns the correct subquery result, 
![image](https://github.com/AlaSQL/alasql/assets/31096082/3b5eb8b7-52cd-4e96-b347-438ca8f5cd53)


there's no procedure following this that would join this specific result back with the exisiting result set. 

One possible solution could be to update the source.data with the result of subquery and add the following procedures that's used in queries of type "queryparam"
![image](https://github.com/AlaSQL/alasql/assets/31096082/be5e8138-70f9-475c-807f-74024a25b2ab)

Have had success in joining with results of non aggregate subqueries. 
With aggregates like count, there seems to be some problem, im looking for a solution for the same.

adarshjhaa100:
Found the issue, was accidentally passed the param data rather than query set:
![image](https://github.com/AlaSQL/alasql/assets/31096082/fc5097de-3a5e-4c19-8576-ef24ef69d6cd)

Getting correct join results for the subquery pointed out in the bug report for the following value of x:

[...]

PR #1800, opened 2023-09-16
Fix for #1789: JOINs only return the second subquery

adarshjhaa100:
This PR fixes the issue #1789.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.1.9

Issue #1797, opened 2023-09-15
When using SELECT * all FUNC() are returned as a distinct columns even if they are defined "as `col_name`"


phantasma2983:
If we give it this data and this query:

```
var data = [{a:1,b:1,c:1},{a:1,b:2,c:1},{a:1,b:3,c:1}, {a:2,b:1,c:1}];
var res = alasql('SELECT *, COUNT(a) AS d FROM ? GROUP BY a', [data] );

the result is:
[{
  a: 1,
  b: 1,
  c: 1,
  COUNT(a): 3, <--- ???
  d: 3
}, {
  a: 2,
  b: 1,
  c: 1,
  COUNT(a): 1, <--- ???
  d: 1
}]
```

adarshjhaa100:
Hi @mathiasrw would like to work on this if nobody's picked it up yet.

adarshjhaa100:
Hello @mathiasrw found the issue. There's a flaw in how the `selectgfn` method for Query is formed behind the scenes specifically for the current issue. When there's a `select *` in addition to there being column(s) with alias in the query, the `selectgfn` looks something like this:
~~~
(function anonymous(g,params,alasql) {
    var y;
    var r = {};
    for(var k in g) 
    {
        r[k]=g[k]
    };

    r['d']=(y=g['COUNT(a)'],y===y?y:undefined);
    delete r['COUNT(a)']
    return r
})
~~~


Let's consider the above example provided in the issue:

at the stage where `gfn` or `selectgfn` is called for the query in `queryfn3`, 
![image](https://github.com/AlaSQL/alasql/assets/31096082/3449218f-22cb-43ce-a785-8bc3fd7ea695)

the object for `query.group[i]` will be of the format : `{a:"", b:"", c:"", 'COUNT(a)':"" }` . Also, we have an alias for `COUNT(a)` which is `d`. Calling the function above with this data would result in an object of the form `{a:"", b:"", c:"", 'COUNT(a)':"", d:"" }` . 
Clearly this signifies that `Select *` that should pull in all the columns of the data source, rather pulls in all columns of the current state of data which doesn't work in this case. 

Based on my preliminary investigations, here are some of the possible solutions:
1) Find out a way to identify whether a specific column is from the data source (can probably use the keys of of `query.params[i]` or `source.data` to identify this, or could use a flag during initialization of Query object for the same). This solution could have a significant impact and could potentially have some side effects. So, we need to be careful with this.

[...]

PR #1808, opened 2023-09-25
Fix Issue #1797

adarshjhaa100:
**Proposed Change**

Fixed issue #1797 where adding alias alongside select(*) was producing duplicate columns. Added the testcase to the file test1797.js that passed without any issues.

**Testing**
 data and query 
image
Before: output: This has duplicate `COUNT(a)` along with `d`
image

After: output, clearly duplicate col has been removed.
image

After: data and query


mathiasrw:
Released as part of v4.1.10

PR #1822, opened 2023-10-25
Several performance optimizations

paulrutter:
We have been using AlaSQL for a couple of years, and had some patches applied to it locally.
These patches were created after Node.js CPU profiling AlaSQL queries. This PR contains these patches. 

Performance improvements for AlaSQL
- Add AST cache
- Add regex pattern cache when using LIKE (building RegEx is expensive)
- Optimize when having lots of values in a (NOT) IN statement, using JS Set
- Remove RegEx checking for DATE function (what is it used for? It slows DATE functions down)
- When using NOW, just return the date (not sure why a date is translated to a string?)

Please let me know if these changes make sense and/or require changes.
Especially the last two changes might be there for a reason we don't know about.

I also skipped some test cases around string dates and adjusted one for `WHERE IN`, using plain string literals.
The PR is lacking new test cases, but if you're interested in these changes, we can work on these.

I couldn't get the formatter to work on Windows though, so that still has to be done.

### DIFF
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -1212,10 +1212,9 @@ var like = (utils.like = function (pattern, value, escape) {
 		s += '$';
 		//    if(value == undefined) return false;
 		//console.log(s,value,(value||'').search(RegExp(s))>-1);
-		patternCache[pattern] = RegExp(s.toUpperCase())
+		patternCache[pattern] = RegExp(s, 'i');
 	}
-	return ('' + (value || '')).toUpperCase().search(patternCache[pattern]) > -1;
-
+	return ('' + (value ?? '')).search(patternCache[pattern]) > -1;
 });
 
 utils.glob = function (value, pattern) {
--- a/src/15utility.js
+++ b/src/15utility.js
@@ -1,9 +1,9 @@
 /*jshint unused:false*/
 /*
-    Utilities for Alasql.js
+	Utilities for Alasql.js
 
-    @todo Review the list of utilities
-    @todo Find more effective utilities
+	@todo Review the list of utilities
+	@todo Find more effective utilities
 */
 
 /**
@@ -59,7 +59,7 @@ function returnTrue() {
   @function
   @return {undefined} Always undefined
   */
-function returnUndefined() {}
+function returnUndefined() { }
 
 /**
   Escape string
@@ -354,7 +354,7 @@ var loadFile = (utils.loadFile = function (path, asy, success, error) {
 	} else if (utils.isCordova) {
 		/* If Cordova */
 		utils.global.requestFileSystem(LocalFileSystem.PERSISTENT, 0, function (fileSystem) {
-			fileSystem.root.getFile(path, {create: false}, function (fileEntry) {
+			fileSystem.root.getFile(path, { create: false }, function (fileEntry) {
 				fileEntry.file(function (file) {
 					var fileReader = new FileReader();
 					fileReader.onloadend = function (e) {
@@ -565,7 +565,7 @@ var removeFile = (utils.removeFile = function (path, cb) {
 		utils.global.requestFileSystem(LocalFileSystem.PERSISTENT, 0, function (fileSystem) {
 			fileSystem.root.getFile(
 				path,
-				{create: false},
+				{ create: false },
 				function (fileEntry) {
 					fileEntry.remove(cb);
 					cb && cb(); // jshint ignore:line
@@ -633,7 +633,7 @@ var fileExists = (utils.fileExists = function (path, cb) {
 		utils.global.requestFileSystem(LocalFileSystem.PERSISTENT, 0, function (fileSystem) {
 			fileSystem.root.getFile(
 				path,
-				{create: false},
+				{ create: false },
 				function (fileEntry) {
 					cb(true);
 				},
@@ -699,7 +699,7 @@ var saveFile = (utils.saveFile = function (path, data, cb, opts) {
 		} else if (utils.isCordova) {
 			utils.global.requestFileSystem(LocalFileSystem.PERSISTENT, 0, function (fileSystem) {
 				//                alasql.utils.removeFile(path,function(){
-				fileSystem.root.getFile(path, {create: true}, function (fileEntry) {
+				fileSystem.root.getFile(path, { create: true }, function (fileEntry) {
 					fileEntry.createWriter(function (fileWriter) {
 						fileWriter.onwriteend = function () {
 							if (cb) {
@@ -774,7 +774,7 @@ var saveFile = (utils.saveFile = function (path, data, cb, opts) {
 				disableAutoBom: false,
 			};
 			alasql.utils.extend(opt, opts);
-			var blob = new Blob([data], {type: 'text/plain;charset=utf-8'});
+			var blob = new Blob([data], { type: 'text/plain;charset=utf-8' });
 			saveAs(blob, path, opt.disableAutoBom);
 			if (cb) {
 				res = cb(res);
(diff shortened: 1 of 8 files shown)


mathiasrw:
I love it!

Let me have a look at the code and merge as soon as possible

paulrutter:
I was under the impression that the failing unit tests where due to running on Windows (newlines), but apparantly these also fail in CI. Are these new failures? Then we need to look into these.

paulrutter:
@mathiasrw I fixed the unit tests for the LIKE, all unit tests succeed now locally:

```
  ............................................................................................................................................................
  ....,,,......................................,,,................,...........................................................................................
  .............................................................................................................................,,..,................,.........
  ...............................,.........................................................................................................................,,,
  ,,,,............................,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,..........................................,,,...................,,,,......,,,,........
  ..,,,,,,,,,,,,,,,...,,,...,,,,,,,,,,,,,,,,,,.........,,,,,,,,..................................,,,,,,,,,,,,........,,,..............................,,,,,,,,
  ,,.................................,,,,....,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,....,,,,,,................,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
  ,,,,,,,,,,,,,,,,,,,,,,,,,,,....,,,,,,,,.......,,,,,,..........................................,,,,,,,,,,,,,,,,,,,.........,,,,.,............................
  ...........................................................,........,,.,.,..,,,,,,,,,,......................................................................
  .......,.................................,,,,,,,,,,,,,,,,,,,,,,,,,..........................................................................................
  .....................................................,........................................................,.......,,..,,..,,..,,.........,..............
  ..,.......,..,......,,...................,..,........,,,.....,,,,,,,,,,,,,,,,...............,.,.............................................................
  ...................................................,.................

  1561 passing (4s)
  380 pending
```

paulrutter:
@mathiasrw I adjusted the skipped tests and re-enabled them again, as i made it configurable as option instead.
I extended the relevant unit tests.

Can you run the workflow again?

mathiasrw:
I sure can. Sorry for slow response. Lots of things going on at the moment. 

I love the performance optimizations, but we cant "just" redo tests that beforehand was working. If we will be doing changes that by some can be considered breaking (= tests are not verifying the same thing as before) than I want to pool those together and add them at the next major version bump (we got a few other tings that is also in the queue for that)

Are you OK to split this PR into so we seperate what demands a change of an existing test?

test/test811.js
```diff
@@ -102,7 +102,7 @@ describe('Test 811 - String / Number objects', function () {
 	});
 
 	it('5. Where In', function (done) {
-		var t1 = [{ID: new String('s1')}, {ID: new String('s2')}, {ID: new String('s3')}];
+		var t1 = [{ID: "s1"}, {ID: "s2"}, {ID: "s3"}];
```
mathiasrw, line 117: This is a breaking change. We need to seperate that out of this PR

mathiasrw (changes requested):

test/test811.js
paulrutter: I will look into it, maybe the implementation using sets can be adjusted to also allow for using new String.

paulrutter:
@mathiasrw I fixed the implementation so it supports both string literals and new String.
Added new unit tests for using string literals in `IN`, and also added tests using `NOT IN`.

Can you check and run the workflow again? This way, we don't need to split of the PR.

src/61date.js
```diff
@@ -43,27 +43,30 @@ stdfn.OBJECT_ID = function (objid) {
 };
 
 stdfn.DATE = function (d) {
-	if (/\d{8}/.test(d)) return new Date(+d.substr(0, 4), +d.substr(4, 2) - 1, +d.substr(6, 2));
+	if (!isNaN(d) && d.length === 8) return new Date(+d.substr(0, 4), +d.substr(4, 2) - 1, +d.substr(6, 2));
```
mathiasrw, line 46: Good call

src/50expression.js
```diff
@@ -506,8 +509,17 @@
 					s += '.indexOf(';
 					s += 'alasql.utils.getValueOf(' + leftJS() + '))<0)';
 				} else if (Array.isArray(this.right)) {
-					s = '([' + this.right.map(ref).join(',') + '].indexOf(';
-					s += 'alasql.utils.getValueOf(' + leftJS() + '))<0)';
+					// Added patch to have a better performance for when you have a lot of entries in a NOT IN statement
```
mathiasrw, line 512: I love it

src/50expression.js
```diff
@@ -483,14 +483,17 @@
...
 				} else if (Array.isArray(this.right)) {
-					//			if(this.right.length == 0) return 'false';
-					s =
-						'([' +
-						this.right.map(ref).join(',') +
-						'].indexOf(alasql.utils.getValueOf(' +
-						leftJS() +
-						'))>-1)';
-					//console.log(s);
+					// Added patch to have a better performance for when you have a lot of entries in an IN statement
```
mathiasrw, line 486: Nice

src/50expression.js
```diff
@@ -483,14 +483,17 @@
...
+					if (!alasql.sets) {
+						alasql.sets = {};
+					}
+					const allValues = this.right.map((value) => value.value);
+					const allValuesStr = allValues.join(",");
+					if (!alasql.sets[allValuesStr]) {
+						// leverage JS Set, which is faster for lookups than arrays
```
mathiasrw, line 493: Neat

src/20database.js
```diff
@@ -67,6 +67,7 @@ var Database = (alasql.Database = function (databaseid) {
 Database.prototype.resetSqlCache = function () {
 	this.sqlCache = {}; // Cache for compiled SQL statements
 	this.sqlCacheSize = 0;
+	this.astCache = {}; // Cache for AST objects
```
mathiasrw, line 70: Makes sense

src/17alasql.js
```diff
@@ -246,19 +249,28 @@ alasql.dexec = function (databaseid, sql, params, cb, scope) {
...
 		// If database structure was not changed since last time return cache
 		if (statement && db.dbversion === statement.dbversion) {
 			return statement(params, cb);
 		}
 	}
 
-	// Create AST
-	var ast = alasql.parse(sql);
+	var ast = db.astCache[hh];
+	if (alasql.options.cache && !ast) {
```
mathiasrw, line 264: Exactly how I hoped it would be

src/17alasql.js
```diff
@@ -119,12 +119,15 @@ alasql.options = {
...
-	excel: {cellDates: true},
+	excel: { cellDates: true },
 
 	/** Option for SELECT * FROM a,b */
 	joinstar: 'overwrite',
 
 	loopbreak: 100000,
+
+	/** Whether GETDATE() and NOW() return dates as string. If false, then a Date object is returned */
+	dateAsString: true,
```
mathiasrw, line 130: Nice way to default to the old way but keep the ability to run perfomant

src/15utility.js
```diff
@@ -1175,43 +1175,46 @@ var domEmptyChildren = (utils.domEmptyChildren = function (container) {
   @parameter {string} escape Escape character (optional)
   @return {boolean} If value LIKE pattern ESCAPE escape
   */
-
+var patternCache = {};
```
mathiasrw, line 1178: Such a good idea. Love it.

mathiasrw (approved):

mathiasrw (approved):

paulrutter:
Thanks for merging the PR, will this be part of next release?

Issue #1827, opened 2023-11-09
Documentation update for options 


paulrutter:
While working on recent performance improvements, i noticed that the options available in alasql don't seem to be documented anywhere. 

Because some of these impact the performance, like the new `dateAsString`, i think it would make sense to document some of these options, and how to set them when using Alasql. 

If you agree, i could make a PR with some of these. 
Where would be the best place to do this, the README?

paulrutter:
Ah i see https://github.com/AlaSQL/alasql/wiki/AlaSQL-Options, which i can extend and maybe link to in the README?

mathiasrw:
Thanks for the PR. This is exactly how we do better together.

PR #1828, opened 2023-11-09
Add options to readme

paulrutter:
See 
https://github.com/AlaSQL/alasql/issues/1827.

The options wiki page should also be updated with the `dateAsString` option, but i cannot do that via a PR.

Thanks!


mathiasrw:
Love it. Thank you!

Issue #1829, opened 2023-11-10
4.2.0 regression : basic query using 'in' keyword working 4.1.11 and returning no result in 4.2.0


jbrazeau:
I've upgraded alasql (4.1.10 -> 4.2.0) this morning and faced a regression.

I've isolated the problem in a trivial test available in the following git repository : https://gitlab.com/jean-francois.brazeau/alasql-bug-4.2.0

The repository contains a very basic test that :
* creates a simple table (2 columns)
* insert two rows
* selects the rows using `in` keyword

Simply run : 
* `go-4.1.11.sh` to install `alasql@4.1.11` and successfully run the test.
* `go-4.2.0.sh` to install `alasql@4.2.0` and notice that the query returns no result at the end of the script.

paulrutter:
Performance improvements yesterday were indeed affecting IN and NOT IN, so it seems something was missed in the unit test coverage.
I did add a similar test though, so not sure what causes it yet. I might investigate next week.

Probably related to using `?` as using direct data assignments worked in the unit test.

paulrutter:
@jbrazeau i have been able to reproduce and confirmed that my PR is indeed causing the regression.
I will let you know once i have a fix.

paulrutter:
@jbrazeau see related PR which should fix it.

jbrazeau:
Thank you very much for paying attention to that issue.

I've checked out the PR, built alasql and manually "patched" `dist/alasql.fs.js` in my `node_modules` folder (after having installed `alasql@4.2.0`). I confirm that the regression seems to disappears ! 👍

paulrutter:
I would propose merging this and release as 4.2.1.
We can check to re-add the cache layer for the Set later, now we have added more unit test coverage.

mathiasrw:
Released as part of 4.2.1

PR #1830, opened 2023-11-10
Fix regression WHERE (not) in (4.2.0)

paulrutter:
Fix for https://github.com/AlaSQL/alasql/issues/1829.

- Fix regression when using IN / NOT IN
- Using refs didn't work anymore
- Dropped set cache, as it's not really needed unless the same queries are performed often
- Added unit test to guard behavior


paulrutter:
@mathiasrw Some regression introduced in 4.2.0, this PR fixes it.
I removed the Set cache as i couldn't get it to work with passing references. It would still be beneficial when running the same query with lots of IN values against different data though, but we might add that back in later. 

See the new unit test.

mathiasrw:
Thank you so much for acting on this rapidly.

test/test1829.js
mathiasrw, line 1: Absolutely love it. Thank you!

mathiasrw:
Released as part of 4.2.1

PR #1834, opened 2023-11-13
Add Set cache to using WHERE (NOT) IN

paulrutter:
- Add Set cache again, but switch on the "right" having ParamValue instances
- If so, don't use cache (or when cache is disabled)
- If not, use Set cache which will make doing identical (WHERE (NOT) IN) queries on different data faster


mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.2.2

PR #1835, opened 2023-11-13
Fix for SELECT query (a AS b, b AS c), change in behavior between 4.1.9 and 4.1.10

### LINKED ISSUE #1820: SELECT query (a AS b, b AS c), change in behavior between 4.1.9 and 4.1.10
With a SELECT query like the following:

```js
let item1 = { a: 1, b: "hello" };
let item2 = { a: 2, b: "" };

var res = alasql('SELECT a as b, b as c FROM ? GROUP BY a,b', [[item1, item2]]);
console.log(res);
```

I get the following results using `4.1.9` (https://jsfiddle.net/02c9esvw/37/):

```js
[{
  b: 1,
  c: "hello"
}, {
  b: 2,
  c: ""
}]
```

And the following with `4.1.10` (https://jsfiddle.net/kqjtoapz/99/):

```js
[{
  c: "hello"
}, {
  c: ""
}]
```

The same goes when selecting MATRIX like this: `[0] AS [1], [1] AS [2]` which is what we're actually using in some cases.

I would expect the 4.1.9 result is correct. If it's not is this an intentional change that we should adapt to?

paulrutter:
Fixes https://github.com/AlaSQL/alasql/issues/1820

- Only remove columns where the alias is not a column in the result itself
- Add unit test

### DIFF
--- a/test/test1820.js
+++ b/test/test1820.js
@@ -1,0 +1,26 @@
+if (typeof exports === 'object') {
+	var assert = require('assert');
+	var alasql = require('..');
+} else {
+	__dirname = '.';
+}
+
+describe('Test 1820 - SELECT query (a AS b, b AS c)', function () {
+
+	it('1. Where IN with refs', function (done) {
+		let item1 = { a: 1, b: "hello" };
+		let item2 = { a: 2, b: "" };
+
+		var res = alasql('SELECT a as b, b as c FROM ? GROUP BY a,b', [[item1, item2]]);
+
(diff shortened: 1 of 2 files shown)


test/test1820.js
mathiasrw, line 1: Yes!

mathiasrw:
Epic. Thank you so much!

mathiasrw:
Released as part of v4.2.2

PR #1855, opened 2023-12-27
feat: add database and table types for TypeScript

sytone:
This change extends the types already present to allow referencing of the database and tables via TypeScript without type errors being shown. It also makes the validations during compilation and IDE based development better.


mathiasrw:
THank you!

Issue #1180, opened 2020-04-16
Property 'Database' does not exist on type 'AlaSQL'.


mateja176:
AlaSQL is based on unpaid voluntary work. Thank you for taking the time to make it better. 

Question about how to ...
- Please use http://stackoverflow.com/questions/ask?tags=alasql

Something is not working as expected:
- Describe the problem
  - TypeScript type error is thrown based on the lack of a type signature
- Provide code that replicates the problem 
  - https://stackblitz.com/edit/alasql-database-does-no-exist?file=index.ts

rendall:
I address this with PR #1865

mathiasrw:
Fix released as part of v4.2.6

PR #1865, opened 2024-01-26
Add Database interface to type definitions

rendall:
This fixes issue #1180 and adds the property Database to the [types/alasql.d.ts](https://github.com/AlaSQL/alasql/blob/674b9271c1d3c0498e6ce0f6b1f53baa44671c25/types/alasql.d.ts) file.

The [Getting Started](https://github.com/alasql/alasql/wiki/Getting%20started) section has this code:

```JavaScript
    var mybase = new alasql.Database();
    mybase.exec('CREATE TABLE one (two INT)');
```

However, when using TypeScript, this causes a type error such as:
```bash
src/lib/utils.ts:239:25 - error TS2551: Property 'Database' does not exist on type 'AlaSQL'. Did you mean 'databases'?

239   const db = new alasql.Database();
                            ~~~~~~~~
```

The cause is that [types/alasql.d.ts] has no `Database` in its type definitions.  This PR adds that type definition.

### DIFF
--- a/types/alasql.d.ts
+++ b/types/alasql.d.ts
@@ -138,6 +138,22 @@ declare module 'alasql' {
 		[tableName: string]: table;
 	}
 
+	interface Database {
+		databaseid: string;
+		dbversion: number;
+		tables: {[key: string]: any};
+		views: {[key: string]: any};
+		triggers: {[key: string]: any};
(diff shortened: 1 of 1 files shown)


mathiasrw:
Thank you!

Released as part of v4.2.6

PR #1872, opened 2024-02-02
Fix #1109

### LINKED ISSUE #1109: Uncaught TypeError: Cannot read property 'length' of undefined while exporting to Excel
While exporting the excel file i get the error saying "Uncaught TypeError: Cannot read property 'length' of undefined".
The scenario is : i have 6 different table which contains different data.When i am going to export it to excel sheet, it should create different 6 sheets in the Excel workbook.
When i have 1 empty table having no data, it creates an excel workbook with 1 empty sheet/tab but when i have more than 1 empty table, it fails to create an excel sheet and throws an error.

here is my query for the same.

var opts = [{sheetid:"Operations",header:true},{sheetid:"Operation Hooks",header:true},{sheetid:"Tool Groups",header:true},{sheetid:"BuyOffs",header:true},{sheetid:"Data Collection Groups",header:true},{sheetid:"DC Params",header:true}];
  		alasql('SELECT INTO XLSX(" ' + saveFile + '.xlsx", ?) FROM ?', [opts,[data,data1,data2,data3,data4,data5]]);

where data,data1,data2,data3,data4,data5 are the the data model(array) having the table data.

can someone please look into this. it needs to resolve in an urgency.

gkkachi:

### DIFF
--- a/src/833xlsx.js
+++ b/src/833xlsx.js
@@ -90,10 +90,14 @@ alasql.into.XLSX = function (filename, opts, data, columns, cb) {
 		var dataLength = Object.keys(data).length;
 
 		// Generate columns if they are not defined
-		if ((!columns || columns.length == 0) && dataLength > 0) {
-			columns = Object.keys(data[0]).map(function (columnid) {
-				return {columnid: columnid};
-			});
+		if (!columns || columns.length == 0) {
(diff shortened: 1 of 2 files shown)


mathiasrw:
Thank you!

PR #1900, opened 2024-03-16
fix(group_concat):  Added fix for group_concat #1898

### LINKED ISSUE #1898: group_concat function doesn't work in lower case
when used as `GROUP_CONCAT` in query it works but when used `group_concat` it throws the following:

```
TypeError: alasql.fn["group_concat"] is not a function. (In 'alasql.fn["group_concat"](g['fieldName'])', 'alasql.fn["group_concat"]' is undefined)
```

this is not the case with other aggregation functions like `count` for example.

vishal6557:
Fix #1898 

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -256,7 +256,7 @@ stdfn.CONCAT_WS = function () {
 // TRIM
 
 // Aggregator for joining strings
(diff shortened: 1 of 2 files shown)


vishal6557:
@mathiasrw  Please review it. Thanks!

mathiasrw:
Thank you!

PR #1903, opened 2024-03-17
Invalid query throws n.Term #1871

### LINKED ISSUE #1871: Invalid query throws `n.Term is not a constructor`
I've noticed that the error message on the invalid token looks like this:

> n.Term is not a constructor

When executing:

```
> xxxx
```

It looks like the invalid token is not handled and it just throws an unhandled exception.

You can see this on my Terminal App:

https://codepen.io/jcubic/pen/dVBaRm

vishal6557:
Fix #1871 

### DIFF
--- a/src/alasqlparser.jison
+++ b/src/alasqlparser.jison
@@ -2960,12 +2960,11 @@ TermsList
 	;
 
 Term
-	: Literal
-		{ $$ = new yy.Term({termid:$1}); }
-	| Literal LPAR TermsList RPAR
-		{ $$ = new yy.Term({termid:$1,args:$3}); }
-	;
-
+    : Literal
+        { $$ = {termid: $1}; }
+    | Literal LPAR TermsList RPAR
+        { $$ = {termid:$1, args:$3}; }
+    ;
 Query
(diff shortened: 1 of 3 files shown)


vishal6557:
@mathiasrw  Have a look, any feedback is appreciated!

mathiasrw:
Very interesting how you solved this. Im wondering why the `yy.Term` was there in the first place. 

Ill merge it. 

Did you have any problem generating the file from the jison? I had some challenges with deep dependencies lately. Or did you just update the actual .js file?

mathiasrw:
Fix released as part of v4.3.1

vishal6557:
@mathiasrw No, I don't have a problem generating files from the `jison`

I think `yarn jison` command is broken, I manually install `jison` globally and using command 
`jison ./src/alasqlparser.jison -o ./src/alasqlparser.js`

Can you try this command if it works, I'll create a pull request and update the `package.json`

mathiasrw:
Neat! Ill look into this!

mathiasrw:
It works. Neat!

PR #1919, opened 2024-05-06
Add jsonl support

kfigiela:
JSON support is great, but JSONL is very common in streaming cases (e.g. logs). 

```
$ cat test.jsonl
{"a": 1}
{"a": 2}
{"a": 42}

$ alasql "select * from jsonl('./test') where a > 1"
[
  {
    "a": 2
  },
  {
    "a": 42
  }
]
```

### DIFF
--- a/src/84from.js
+++ b/src/84from.js
@@ -131,28 +131,46 @@ alasql.from.JSON = function (filename, opts, cb, idx, query) {
 		if (cb) {
 			res = cb(res, idx, query);
 		}
-	});
+	}),
+		err => {
+			throw new Error(err);
+		};
 	return res;
 };
 
-alasql.from.JSONL = function (filename, opts, cb, idx, query) {
-	var res;
-	filename = alasql.utils.autoExtFilename(filename, 'jsonl', opts);
-	alasql.utils.loadFile(filename, !!cb, function (data) {
-		res = data.split(/\r?\n/);
-		// Remove last line if empty
-		if (res[res.length - 1] === '') {
-			res.pop();
-		}
-		for (var i = 0, ilen = res.length; i < ilen; i++) {
-			res[i] = JSON.parse(res[i]);
-		}
-		if (cb) {
-			res = cb(res, idx, query);
-		}
-	});
-	return res;
+const jsonl = ext => {
+	return function (filename, opts, cb, idx, query) {
+		let out = [];
+		filename = alasql.utils.autoExtFilename(filename, ext, opts);
+		alasql.utils.loadFile(
+			filename,
+			!!cb,
+			function (data) {
+				data.split(/\r?\n/).forEach((line, ix) => {
+					const trimmed = line.trim();
+					if (trimmed !== '') {
+						// skip empty lines, we do not use filter on an input, as we want to preserve line numbers
+						try {
+							out.push(JSON.parse(trimmed));
+						} catch (e) {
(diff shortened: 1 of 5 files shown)


src/84from.js
```diff
@@ -135,6 +135,25 @@ alasql.from.JSON = function (filename, opts, cb, idx, query) {
 	return res;
 };
 
+alasql.from.JSONL = function (filename, opts, cb, idx, query) {
+	var res;
+	filename = alasql.utils.autoExtFilename(filename, 'jsonl', opts);
+	alasql.utils.loadFile(filename, !!cb, function (data) {
+		res = data.split(/\r?\n/);
+		// Remove last line if empty
+		if (res[res.length - 1] === '') {
```
mathiasrw: It could be more than one line that is empty, or an empty line in the middle. I thin its also good to ignore lines with only space. I propose you add 

.map(line => line.trim()).filter(Boolean); to line 142 and remove line 144-146

src/84from.js
```diff
@@ -135,6 +135,25 @@ alasql.from.JSON = function (filename, opts, cb, idx, query) {
...
+			res.pop();
+		}
+		for (var i = 0, ilen = res.length; i < ilen; i++) {
+			res[i] = JSON.parse(res[i]);
```
mathiasrw: Are we happy with a random error message, or should we provide more informative error message from a try-catch?

mathiasrw:
Love it. Good and solid functionality to be added. Thank you. 

Any chance you can add a test file that confirms its working (so we detect if we break that later)?

Is `.jsonl` the right name? What about `.ndjson` (new line delimited json) or I have even seen `.jsons` ?

mathiasrw (changes requested):

mathiasrw:
@kfigiela ?

kfigiela:
@mathiasrw, thanks for reminder. I'll try to address those over the weekend and add tests.

As far as the extension/naming, according to [a blog post](https://www.hypertesto.me/en/blog/jsonl-and-ndjson/) these are "competing", but practically identical standards. We only parse, so as long as JSON.parse can handle it we should be good. I think, it makes sense to register them both, so either file extension works.

src/84from.js
kfigiela: Right, enriching with line number can be super useful in case of invalid input.

paulrutter:
https://duckdb.org/2023/03/03/json.html uses ndjson indeed, would be good to rename.

kfigiela:
On the other hand, Google BigQuery mentions JSONL everywhere. Nevertheless, we can have both aliases working and this is what I have implemented. Otherwise, I addressed the above comments and added some tests.

mathiasrw:
Just provide all the most used names (alasql.jsonl = alasql.ndjson = alsql.....)

mathiasrw:
Ahh - you already did :)

src/84from.js
```diff
@@ -131,10 +131,47 @@ alasql.from.JSON = function (filename, opts, cb, idx, query) {
 		if (cb) {
 			res = cb(res, idx, query);
 		}
-	});
+	}),
+		err => {
+			throw new Error(err);
+		};
 	return res;
```
mathiasrw, line 138: THanks

mathiasrw (approved):

mathiasrw:
Doing a bit of performance testing. 

image

its faster just to regex `\s*\r?\n\s*/` as it removes any empty lines in one swoop. Ill add this before merging.

Issue #1936, opened 2024-06-18
Add date function to get current date without time component


paulrutter:
Please introduce a function to get the current date without time component, so this is no longer needed:

```
DATE(YEAR(NOW()) + '-' + MONTH(NOW()) + '-' + DAY(NOW()))
```

See https://www.w3schools.com/sql/func_mysql_curdate.asp for example.

paulrutter:
https://github.com/AlaSQL/alasql/pull/1937

paulrutter:
Naming of the function is up for debate.

paulrutter:
https://github.com/AlaSQL/alasql/blob/ca7ecb4d38a5810d22d05ea7eac7f60f7f10ba75/src/alasqlparser.jison#L110 probably needs updating as well.

paulrutter:
Fixed in https://github.com/AlaSQL/alasql/pull/1937

PR #1937, opened 2024-06-18
Add date function to get current date without time component

paulrutter:
- Add CURDATE/CURRENT_DATE method to AlaSQL
- Add test

This removed the need for:

```
DATE(YEAR(NOW()) + '-' + MONTH(NOW()) + '-' + DAY(NOW()))
```


paulrutter:
@mathiasrw any feedback on this pr? Thanks!

mathiasrw:
Interesting, relevant and missing. Nice. 

    SELECT CURDATE();
    
looks good to me. I do feel like we should avoid none standard notation like `SELECT CURRENT_DATE();` and instead support the widely use same notation as MySQL and MariaDB with `SELECT CURRENT_DATE;` where its not a function, but a constant. 

Are you feeling the mojo to try to dive into that? Adding it to src/alasqlparser.jison should not be too complex, but I admit its a bit exotic if you have not worked with parsers before.

paulrutter:
Isn't the function exactly what MySQL offers?
See https://www.w3schools.com/sql/func_mysql_curdate.asp and https://www.w3schools.com/sql/func_mysql_current_date.asp and https://www.techonthenet.com/mariadb/functions/current_date.php#:~:text=The%20CURRENT_DATE%20function%20will%20return,synonym%20for%20the%20CURDATE%20function.

I do see that most also support an alias as constant, although we don't offer this for other date functions except CURRENT_TIMESTAMP in alasql. The current approach would be in line with other AlaSQL date functions. Do you want me to look into using these aliases? 

To the parser point, i'm not seeing the other date functions in there. Is that deliberate? Only CURRENT_TIMESTAMP is there, not NOW for example.

I can look into adding it to the parser though next week.

paulrutter:
@mathiasrw I tried rebuilding the parser via `npm run jison`. but the resulting alasqlparser.js is completely different from what's in git right now. Should i use a different command?

I wouldn't expect changes against what's in git.
Also, when i run the tests after generating the parser again, a lot of tests fail due to the parser throwing errors.

I found https://github.com/AlaSQL/alasql/pull/1903#issuecomment-2004332511, will try that and update the package.json if that works.

paulrutter:
@mathiasrw 
Can you take a look at the changes made?
I also added a section to rebuild the parser and added both CURDATE as CURRENT_DATE.
It can be used with and without parenthesis.

It feels to me like other date functions should be added to the parser as well, as these are currently missing.
`DATE`/`NOW`/`GETDATE`/`YEAR`/`MONTH` for example.

paulrutter:
https://github.com/AlaSQL/alasql/issues/1936

README.md
```diff
@@ -629,6 +629,15 @@ AlaSQL is an [OPEN Open Source Project](http://openopensource.org/). This means
...
 
+## Rebuilding the parser 
+
+To rebuild the parser, follow these steps:
+
+* Make changes to alasqlparser.jison
+* `npm install -g jison`
+* `npm run jison`
+* `npm test` to validate the changes made
+* Commit changes to alasqlparser.jison and alasqlparser.js
```
mathiasrw, line 640: Thank you!

package.json
```diff
@@ -62,7 +62,7 @@
 		"eslint-plugin-promise": "6.1.1",
 		"git-branch-is": "4.0.0",
 		"husky": "8.0.3",
-		"jison-gho": "^0.6.1-216",
```
mathiasrw, line 65: Good to see you dived in. There was a security issue with a subdependency - that is why we transitioned to -gho. Looks like its fixed now.

package.json
```diff
@@ -15,15 +15,15 @@
 	"scripts": {
 		"test": "sh build.sh && yarn test-only",
 		"test-ci": "(yarn test-format || 1) && yarn test && yarn install-g && alasql 'select 1 as Succes'",
-		"test-only": "mocha ./test --reporter dot # npx bun node_modules/.bin/mocha ./test --reporter dot",
+		"test-only": "node node_modules/mocha/bin/mocha.js ./test --reporter dot",
 		"#test-only": "(command -v bun && bun node_modules/.bin/mocha ./test --reporter dot) || npx bun node_modules/.bin/mocha ./test --reporter dot",
```
mathiasrw, line 19: Thanks!

mathiasrw (approved):

paulrutter:
Thanks for merging, can we expect a new release soon?
@mathiasrw

mathiasrw:
Released as part of v4.5.0

Issue #1904, opened 2024-03-21
EXISTS does not work with subqueries


alsundukov:
The EXISTS function works correctly in the main query but throws an error when used in a subquery. For example:

```
res = alasql('SELECT EXISTS(SELECT "a"), a FROM ?', [[item1, item2]]);
// returns expected value

res = alasql('SELECT * FROM (SELECT EXISTS(SELECT "a"), a FROM ?)', [[item1, item2]]);
// error: Cannot read properties of undefined (reading '0')
```

Additionally, I've noticed that the function also fails when used with SET:

`SET @ress = (SELECT EXISTS(SELECT "test"));`

mathiasrw:
Released as part of v4.5.0

alsundukov:
Unfortunately, [this ](https://github.com/AlaSQL/alasql/commit/aeb3b00c926dcc7e7ea7c7e15c95323261c66f5d) does not fix the problem in any way. Now exists returns a Boolean value (true/false) instead of a number (0.1), but this still doesn't work for subquery(((. 

Why wasn't the test added?

PR #1939, opened 2024-06-28
Fix EXISTS Function for Subqueries (Issue #1904)

Midwayne:
Changes Made:
- Updated yy.ExistsValue class in 41exists.js to correctly handle EXISTS function in subqueries.
- Adjusted toJS method to return a boolean value based on the presence of data in the subquery result.

Testing:
- All tests passed successfully when executed with yarn test.

Related Issue:
This pull request addresses [issue #1904](https://github.com/AlaSQL/alasql/issues/1904).

### DIFF
--- a/src/41exists.js
+++ b/src/41exists.js
@@ -1,8 +1,9 @@
 /*
 //
-// EXISTS and other subqueries functions  functions for Alasql.js
+// EXISTS and other subqueries functions for AlaSQL.js
 // Date: 03.11.2014
 // (c) 2014, Andrey Gershun
+// Modified by: Midwayne
 //
 */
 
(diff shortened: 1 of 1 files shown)


src/41exists.js
```diff
@@ -20,12 +21,13 @@ yy.ExistsValue = class ExistsValue {
 	}
 
 	toJS(context, tableid, defcols) {
-		return `this.existsfn[${this.existsidx}](params, null, ${context}).data.length`;
+		// Updated to return a boolean value correctly
+		return `!!this.existsfn[${this.existsidx}](params, null, ${context}).data.length`;
```
mathiasrw, line 25: Oh - it was just the `!!` that was an issue? Amazing

mathiasrw (approved):
Any chance you can add a test file for this issue so we know if we

mathiasrw:
Released as part of v4.5.0

Issue #1684, opened 2023-05-10
UNION ALL still not returning correct results


gregoryAppcraft:
This is a follow-up for https://github.com/AlaSQL/alasql/issues/1415 and uses version 3.1.0 which fixes some issues but not all.

## Describe the problem

UNION ALL fails most of the time depending on how the different SELECT queries are written.

Given the following data : 
```
const data = 	[
		{ "city": "Madrid", "population": 3041579 },
		{ "city": "Rome", "population": 2863223 },
		{ "city": "Paris", "population": 2249975 }
]
```
This works: 

```
var sql =   "SELECT city FROM :data WHERE city = 'Madrid' \
UNION ALL SELECT city FROM :data WHERE city = 'Rome' \
UNION ALL SELECT city FROM :data WHERE city = 'Paris' \
"
var res = alasql(sql, { data });
```

But this doesn't work 
```
var sql =   "SELECT * FROM :data WHERE city = 'Madrid' \
UNION ALL SELECT * FROM :data WHERE city = 'Rome' \
UNION ALL SELECT * FROM :data WHERE city = 'Paris' \
"
var res = alasql(sql, { data });
```

and just returns 
```
[
	{
		"city": "Madrid",
		"population": 3041579
	},
	{},
	{}
]
```

When using `count(*)` or other aggregations, the first select works, but the unioned select doesn't.

## Provide code that replicates the problem

http://jsfiddle.net/GregoryAppCraft/vLjmfeq6/3/

I took a quick look at https://github.com/AlaSQL/alasql/pull/1629 and could try adding more tests and see if I can get it sorted. 

Any advice ? :)

mathiasrw:
Thank you so much for looking into this. 

I think we have to "translate" the `*` into a specific list of fields in order to solve this. 

If you can look at #1629 that would be amazing.

efsouza21:
You can use UNION ALL CORRESPONDING instead of just UNION ALL. That would be a workaround at least here.

mathiasrw:
Released as part of v4.5.0

PR #1940, opened 2024-07-02
fix insertion of empty objects with union all statements

kmr-rohit:
This is an attempt to fix the issue #1684 

Considering approach from #1415 , implemented similar empty query conditional check for UNION ALL 

will be actively looking out for better approach for the problem.


mathiasrw:
Thank you!

mathiasrw:
Released as part of v4.5.0

Issue #1979, opened 2024-10-04
Add GETDATE, NOW as alias to parser + allow CURRENT_TIMESTAMP with parenthesis


paulrutter:
We're in the process of using the AlaSQL parser to validate query syntax in our product.
The AlaSQL parser currently does not recognize the following aliases to `CURRENT_TIMESTAMP` (see `61date.js`).

- `NOW`
- `GETDATE`

Also, `CURRENT_TIMESTAMP` is recognized by the parser, while `CURRENT_TIMESTAMP()` is not.
Do note that all of these functions _do_ work in a query, the parser just cannot properly validate their use.

I'm going to supply a PR to get this fixed, which has one drawback: `now`, `current_timestamp` and `getdate` can no longer be used as column aliases, as they're now reserved keywords.

Let me know what you think @mathiasrw.

paulrutter:
Ps. The same applies to other date functions like DAY() and YEAR(), but that would break far more queries (although it would be more consistent). I tried that locally a while ago. See https://github.com/AlaSQL/alasql/pull/1937#issuecomment-2198488165

With the proposed PR, at least all functions for getting the current date (+time) are consistently handled.

mathiasrw:
Thanks so much for contributing! I'll merge this when we're ready to bump the major version. Since we're currently restructuring the code to be ES6-friendly, it might take a few months before we can release breaking changes.

mathiasrw:
Is this a duplicate of https://github.com/AlaSQL/alasql/pull/1981?

paulrutter:
> Is this a duplicate of #1981?

No, that seems entirely unrelated. Did you mean https://github.com/AlaSQL/alasql/pull/1982?

https://github.com/AlaSQL/alasql/pull/1982 indeed builds on this PR.
We could close https://github.com/AlaSQL/alasql/pull/1980 and just go for #1982 instead, as it adds far more date functions to the parser.

PR #1980, opened 2024-10-04
Add GETDATE, NOW as alias to parser + allow CURRENT_TIMESTAMP with parenthesis

paulrutter:
Add the following functions to the parser:
- NOW
- GETDATE

Also add `CURRENT_TIMESTAMP()` to the parser (with parenthesis).

- Adjust failing test as `now` is a reserved keyword. It was already, it was just not validated.
- Add test

See #1979 


paulrutter:
@mathiasrw whenever you have time, could you please check out this PR? Thanks.

mathiasrw:
Hi Paul.  I _really_ like what you did here. Thank you so much. This is probably how we should have done it from the beginning. 

The changes in [test/test845.js](https://github.com/AlaSQL/alasql/pull/1980/files#diff-9c3bce042e0e355f3ef04987c07438ad14f7b3099064177ac0b0484daf553b46) indicate that this is a breaking change so I will bundle this into a bucket of changes waiting for us to bump major version. Probably during the next few months as we are in the work of restructuring the whole source into ES6 friendly components.

paulrutter:
@mathiasrw Thanks, good to know that will take a while for this to get into a release.
Since we're making changes on this front, it might be worthwhile to also add all other date functions in the same matter (see https://github.com/AlaSQL/alasql/pull/1937#issuecomment-2198488165).

Do you agree? If so, i could work on a PR (WiP: https://github.com/AlaSQL/alasql/compare/develop...blueconic:alasql:feature/add-all-date-functions-to-parser?expand=1)

mathiasrw:
@paulrutter We are talking about all fields named DATE/NOW/GETDATE/YEAR/MONTH having to be wrapped in escape chars, right?

paulrutter:
@mathiasrw yes, indeed. In backticks/quotes.

paulrutter:
@mathiasrw See https://github.com/AlaSQL/alasql/pull/1982 for additional work on the parser.
One can still use `YEAR`, `MONTH` etc without escape chars in `DATEADD` for example, although one cannot use it as column alias without escape characters.

paulrutter:
Closed in favor of #1982

mathiasrw:
Sooooo, this one is to be closed?

paulrutter:
@mathiasrw this was more the issue related to the original PR (#1980), which i closed.
We can close this issue, but then we also have the new PR #1982 still open (which is fine to me).

PR #1990, opened 2024-12-13
Fix Bad Callback For Error

jpulec:
### Description
This callback isn't always passed, since these functions all support promises. If the callback is not defined, the error should just be re-thrown.


mathiasrw:
Spot on. Thank you!

PR #2028, opened 2025-01-08
Remove `data` from the stored `sqlCache` statement closure

### LINKED ISSUE #2027: `sqlCache` retains data (and other properties), possible memory leak
We recently observed an out of memory situation with AlaSQL 4.6.0.
It seems like the `sqlCache` (which is enabled by default), retains more information than it should.

See the attached screenshot of the heapdump, where it shows the `sqlCache` retaining properties like `data`, but also other information.

I doubt that this information should be stored in the cache, as it makes the cache much larger than required.
As a workaround i can either disable the cache, or reduce the `MAXSQLCACHESIZE`, but i'm not sure of the performance implications of doing this.

It seems like the objects stored in the cache should be stripped of any query execution specific information.

image
image

@mathiasrw what do you think?

paulrutter:
Remove the `data` property from the sql cache after executing the query, as it's pointless for re-using from cache.
It just retains table data which could otherwise be garbage collected sooner.
Fixes #2027.

### DIFF
--- a/test/test2027.js
+++ b/test/test2027.js
@@ -22,6 +22,12 @@ describe('Test 2007 - SQL cache', function () {
 		assert.deepEqual(alasql.databases["test"].sqlCache["-169125189"].query.data, []);
 		assert.equal(res.length, 3);
 
+		// Delete all rows
+		alasql('DELETE FROM osoby');
+
+		// Assert that the cache is still empty for "data"
+		assert.deepEqual(alasql.databases["test"].sqlCache["-169125189"].query.data, []);
+
 		// Insert more rows
 		alasql('INSERT INTO osoby VALUES (4, "Jack"), (5, "Paul")');
 
@@ -30,6 +36,6 @@ describe('Test 2007 - SQL cache', function () {
 
 		// Cache should still be empty for "data"
 		assert.deepEqual(alasql.databases["test"].sqlCache["-169125189"].query.data, []);
-		assert.equal(res2.length, 5);
+		assert.equal(res2.length, 2);
 	});
 });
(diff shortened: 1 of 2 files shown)


test/test2027.js
```diff
@@ -0,0 +1,42 @@
...
+
+		assert.deepEqual(alasql.databases["test"].sqlCache["-169125189"].query.data, []);
+		assert.equal(res.length, 3);
+
+		// Delete all rows
+		alasql('DELETE FROM osoby');
+
+		// Assert that the cache is still empty for "data"
+		// Without the fix, the cache would still contain the data from the previous query even though all rows were deleted
+		assert.deepEqual(alasql.databases["test"].sqlCache["-169125189"].query.data, []);
```
paulrutter, line 30: This is the relevant part; without the fix, rows that are already deleted from the table, are still present in the cached data. Thus, this is a memory leak as data is retained that should already have been garbage collected.

Depending on the number of rows in the table, this can have significant impact on memory usage.

mathiasrw:
Fix released as part of v4.2.6

paulrutter:
@mathiasrw thanks for merging!

Btw, we just released a feature to our product using AlaSQL, see https://support.blueconic.com/en/articles/254081-q1-2025-blueconic-release-notes.

A nice showcase of the power of AlaSQL.

mathiasrw:
This is great. It makes me very happy to see things like this!

And what a grat use case for AlaSQL: timeline rollup specified by the user.

paulrutter:
That's why i have been working on the parser, as we use it for several things in the product.
Extracting certain date constructs, but also to give user feedback upon writing a query.

PR #2051, opened 2025-02-11
better error handling

t83714:
- fixes an issue that the error triggered in alasql.utils.loadFile is returned as result incorrectly
- make the user code possible to capture errors thrown in alasql.utils.loadBinaryFile


mathiasrw:
Thank you!

PR #2113, opened 2025-05-31
fix: utf8 issue

### LINKED ISSUE #2112: Issue with non-ASCII characters in JSON() import
When I create a JSON file with non-ASCII like:

```json
[
  {
    "property": "惺"
  }
]
```

...and import it through the browser into a database.

```js
await alasql.promise([
  'CREATE DATABASE IF NOT EXISTS test1',
  'USE test1',
  'DROP TABLE IF EXISTS testtable',
  'CREATE table testtable',
  'SELECT * INTO testtable FROM JSON("json/test.json")'
]);

const results = await alasql.promise(
  'SELECT * FROM testtable'
);
console.log('results', results[0]);
```

I get `{property: 'æ\x83º'}` as the result instead of what I would expect: `{property: '惺'}`

The same problem does not occur in Node.

The issue is apparently with this function used by `fetchData` which is used by `loadFile` which is used by `alasql.from.JSON`:

```js
function getData(path, success, error) {
	return _fetch(path)
		.then(response => response.arrayBuffer())
		.then(buf => {
			var a = new Uint8Array(buf);
			var b = [...a].map(e => String.fromCharCode(e)).join('');
			success(b);
		})
		.catch(e => {
			if (error) return error(e);
			console.error(e);
			throw e;
		});
}
```

I'm not sure why this is being loaded as an array buffer and then processed in this manner.

When I change the code to the following, the content of my UTF8 JSON file is processed as I would expect:

```js
function getData(path, success, error) {
	return _fetch(path)
		.then(response => response.text())
		.then(txt => {
			success(txt);
		})
		.catch(e => {
			if (error) return error(e);
			console.error(e);
			throw e;
		});
}
```

Can you just change the code as above, or do you need some other handling?

Thanks!

brettz9:

I believe the following safely fixes #2112 .

Adds a utility to fetch binary data specifically, leaving the old function to retrieve regular UTF8 (non-binary) pages.

Changes an existing `JSON()` test to ensure it utilizes the non-binary fetch utility.

May break some existing code if they were relying on non-UTF8-encoded text files, but I think this is something that needs fixing.


brettz9:
Hi... Just wondering if there was anything I could do to clarify or help things along? If just busy, feel free to ignore...

mathiasrw:
Lets rolls this into the next major version update. We cant be breaking existing implementations out of nowhere. 

I am in doubt about the test/test157.json changes. WHy does a smily add a sufficient test? Also, can you please add test for fetchBinaryData both async and sync?

brettz9:
I've amended the commit to add tests. (I've also since realized there was no need for a separate `fetchBinaryData` helper, as the remote fetching here will be asynchronous regardless.)

The reason a smiley adds a sufficient test is because a smiley emoji is a non-ASCII, non-Basic-Multilingual-Plane character, i.e., despite being a single character, it has a string length of two (comprised of two "surrogate" characters). When processed as binary per your preexisting binary fetching code, this character will be garbled, but when read as UTF8 text, it will be preserved.

brettz9:
Is there anything else I can do here? Thanks!

mathiasrw:
Looks great. Thank you!

mathiasrw:
Released as part of v4.7.0

Issue #2149, opened 2025-10-26
Alasql cli docker / pipe issues


ammeek:
Hi! I’m currently working on building a Docker image for AlaSQL and I’ve run into an issue where commands passed to the container aren’t being executed as expected.

From what I’ve investigated so far, it seems that Docker doesn’t attach a TTY by default when running commands through ENTRYPOINT, which changes how the CLI behaves. I tried updating the logic in [alasql-cli.js](https://github.com/AlaSQL/alasql/blob/develop/bin/alasql-cli.js), but I think there may be another underlying issue as well.

Specifically, when piping data into a query, the input is being treated as a SQL statement rather than as data. I can reproduce this behavior both inside and outside of Docker, so it doesn’t appear to be container-specific. Looking at the code in [alasql-cli.js](https://github.com/AlaSQL/alasql/blob/develop/bin/alasql-cli.js), it seems to try to read input directly, whereas I had expected file or stream reading to be handled by the standard FROM helpers such as [alasql.from.TXT](https://github.com/AlaSQL/alasql/blob/0d5c9816659bde946c01a2d6ebb8d062161a1976/src/84from.js#L185), which ultimately call [alasql.utils.loadFile](https://github.com/AlaSQL/alasql/blob/0d5c9816659bde946c01a2d6ebb8d062161a1976/src/15utility.js#L306).

I may be misunderstanding the intended behavior — but it seems like the CLI is handling input differently than the query engine’s built-in file readers. I’d really appreciate any clarification on this, or advice on how best to approach making AlaSQL run reliably inside Docker. Thanks!

## Steps to replicate docker issue:

1. Check out [my fork of the repo](https://github.com/ammeek/alasql)
2. Run the following command to build the alasql docker image:
``` cmd
docker build -t alasql-test .
```
3. run the following command to replicate the issue:
``` cmd
docker run alasql-test "SELECT 1 + 1"
```

Expected output:
``` json
[
  {
    "1 + 1": 2
  }
]
```

Actual output:
```
No SQL to process

AlaSQL command-line utility (version 4.6.6)

Usage: alasql [options] [sql] [params]

Options:
  -v, --version  Echo AlaSQL version                                   [boolean]
  -m, --minify   Minify json output                                    [boolean]
  -f, --file     Load SQL from file                                     [string]
      --ast      Print AST instead of result                            [string]
  -h, --help     Show help                                             [boolean]

Examples:
  alasql "sql-statement"                    Run SQL statement and output result
[...]

PR #2159, opened 2025-11-07
Fix: 2149 cli docker / pipe issues

ammeek:
closes:#2149

This PR fixes an issue where piped input was being treated as SQL statements in the AlaSQL CLI, causing parse errors. It also resolves a related Docker issue where commands passed to the container would not execute correctly due to TTY-dependent logic.

Changes:
- Updated the CLI to let alasql.utils.loadFile handle input streams, ensuring piped data is processed correctly.
- Removed TTY-dependent logic so commands now run reliably in Docker and standard terminals.

These changes allow piped input and Docker commands to work as expected without errors.

### DIFF
--- a/bin/alasql-cli.js
+++ b/bin/alasql-cli.js
@@ -10,7 +10,6 @@
 let alasql = require('../dist/alasql.fs.js');
 let path = require('path');
 let fs = require('fs');
-let stdin = process.openStdin();
 let yargs = require('yargs')
 	.strict()
 	.usage(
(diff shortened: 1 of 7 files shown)


mathiasrw:
Nice very nice. 

But

This behaviour has been there for 9 years

image

We will need to keep the existing behaovur the default and use a flag to do what it was meant to do according to the documentation. 

I will have to think about how we do this smart in terms of having a consistent way of setting the flags AND running the code so we dont break things.

mathiasrw:
Nice. I found a way of having both the legacy AND the documented process working.

mathiasrw:
Released as part of v4.9.0

PR #2287, opened 2025-11-30
Iissue fixed #1740

### LINKED ISSUE #1740: Parse() then AST.toString() doesn't restore square brackets on column name

mathiasrw:


mathiasrw:
@copilot Please make a test file called test/test1740.js and replicate the issue described in #1740 

Then see if it has been fixed with the last commit and if not, please work on fixing it.

mathiasrw:
@copilot 

```
": true,
    "isWebWorker": false
}
/home/runner/_work/alasql/alasql/test


  ........................................................
  ...........................................!

  99 passing (149ms)
  1 failing

  1) Test 046
       FROM as parameter
         FROM array of arrays:

      AssertionError [ERR_ASSERTION]: 15 == 0
      + expected - actual

      -15
      +0
      
      at Context.<anonymous> (test/test046.js:35:11)
      at process.processImmediate (node:internal/timers:483:21)



error Command fai
```

PR #2294, opened 2025-12-02
fix RECORDSET returning empty columns if there's no matched rows issue

mathiasrw:
Please confirm this is truly fixing #1547 make sure there is a solid test file and verify things are nice and tidy.


mathiasrw:
@copilot can you help out here?

PR #2419, opened 2025-12-27
Fix OPEN support in SQL queries

### LINKED ISSUE #2224: Limit the impact of using the keyword: OPEN
Limit the impact of using the keyword: OPEN

Just like in https://github.com/AlaSQL/alasql/pull/2203 we want to limit the impact of having keywords. Preferably by making them keywords only in the context needed so the string/word can be used for things like table names and column names i the data - if possible. It might be too much of a hassle to make the impact less, but lets investigate what is possible - still sticking to the SQL-99 specs. 

Start by making a test file named `test0000.js` where 0000 is the issue ID of this ticket. Replicate the use and mimic what usecases could make sense to test, like as a table name, as a column name, as a graph search. Things lke that. 

The grammar can be updated in alasqlgrammar.jison. Focus on small precise changes. After updating it we need to run `yarn jison && yarn test`

Always run `yarn format` before committing.

ericmort:
Fixes #2224


mathiasrw:
Nice. Thank you.

PR #2424, opened 2026-01-04
Deep strict assert test results

mathiasrw:


mathiasrw:
To better detect chances in returning `null` or `undefined`

PR #2456, opened 2026-03-29
Make `react-native-fs` an optional peer dependency

### LINKED ISSUE #2454: Dependency on outdated version of `glob`
Hi there!

Our organization's Black Duck has started flagging alasql for its dependency on glob v7, which is severely outdated, with the most recent major being v13.

I see there's an open PR to update to the most recent version, is there a chance to get that in?
https://github.com/AlaSQL/alasql/pull/2439

Thank you very much in advance and keep up the good work!

Edit: I just took a look at your `package.json` and you actually have an override for glob@13 in there. However, in my project, alasql pulls in glob@7:
```
    "node_modules/alasql/node_modules/glob": {
      "version": "7.2.3",
```

And I'm getting warnings when installing:
> npm warn deprecated glob@7.2.3: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me

I'll do some further digging to find out where that dependency comes from and report back.

ej612:
So far, `react-native-fs` has been declared under `optionalDependencies`, as alasql will use it if available, but doesn't require it to be there. This relation, however, is that of an optional peer dependency, which is contrary to how `optionalDependencies` works, which is that npm will install it by default. To relieve people from having to use custom logic if they don't want to end up with `react-native-fs`, we turn that package into a peer dependency, and use npm's [peerDependenciesMeta](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta) field to declare it as optional.

Note that this is a breaking change as npm will no longer pull in `react-native-fs` by default.

Fixes: #2454

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,13 @@
 		"cross-fetch": "4.1.0",
 		"yargs": "16"
 	},
-	"optionalDependencies": {
+	"peerDependencies": {
 		"react-native-fs": "^2.20.0"
+	},
+	"peerDependenciesMeta": {
+		"react-native-fs": {
+			"optional": true
+		}
 	},
 	"devDependencies": {
 		"blueimp-md5": "2.19.0",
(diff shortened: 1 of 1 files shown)


mathiasrw:
Oh, nice. Thank you!

And thank you for pointing out that this is a breaking change. We got a few of them in pipeline so will merge as soon as we are ready with the next major release.

PR #2465, opened 2026-04-08
fix: resolve Promise.all concurrent query collision with identical SQ…

### LINKED ISSUE #1953: Promise.all error
```
(async () =>{
        async function fun1(){
            let f1 = "D:\\a1.xlsx"
            const res1 = await alasql.promise(`select * from xlsx(?) limit 100000`, [f1])
            return res1
        }

        async function fun2(){
            let f2 = "D:\\a2.xlsx"
            const res2 = await alasql.promise(`select * from xlsx(?) limit 100001`, [f2])
            return res2
        }

        try{
            const orderData = await Promise.all([fun1(), fun2()])    
            console.log(orderData)
        }catch(error){
            console.error(error);
        }finally{
            console.log('end');
        }
    })() 
```  

I have an issue where I've created two asynchronous functions, fun1 and fun2, and I want to execute them in batch using Promise.all, passing parameters f1 and f2 to the xlsx(?) placeholder.

When running these functions individually, the query results are correct.

However, when I put them into Promise.all, no results are returned, and no errors are thrown.

After repeated testing, I found that if there is nothing after xlsx(?) in Promise.all, no results are returned. But if I add a limit and the number after limit is different for each function, the correct results are obtained. Yet, if the numbers after limit are the same, no results are returned again.

For example, in the code above, if both are limit 1000, no results are returned. But if one is limit 1000 and the other is limit 1001, the correct results can be obtained.

What could be the reason for this?

imsanupm:
## Description

Fixes #1953

When two async functions with identical SQL strings were executed 
concurrently via `Promise.all`, no results were returned and no error 
was thrown.

## Root Cause

Two separate issues were working together to cause this bug:

1. **`cleanupCache` mutating shared cached object** — The compiled 
   statement stored in `sqlCache` is shared across concurrent executions. 
   `cleanupCache` was clearing `statement.query.data = []` on this shared 
   object while another concurrent call was still reading it, silently 
   wiping the results.

2. **Shared params reference** — Concurrent executions via `Promise.all` 
   were sharing the same params reference, allowing one call to overwrite 
   another's parameters (`f1`/`f2`).

## Why it was hard to notice

- Running functions **individually** worked fine
- Running with **different SQL strings** (different LIMIT values) worked 
  fine because they got different cache keys
- **No error was thrown** — results were just silently empty
- Only triggered by **concurrent identical SQL strings** via `Promise.all`

## Changes

- `cleanupCache()` — Removed mutation of `statement.query.data` on the 
  shared cached statement object
- `alasql.dexec()` — Clone params before passing to cached statement 
  execution to isolate each concurrent call's parameters

## How to reproduce (before fix)

```js
async function fun1() {
    return await alasql.promise(`select * from xlsx(?)`, ["D:\\a1.xlsx"]);
}
async function fun2() {
    return await alasql.promise(`select * from xlsx(?)`, ["D:\\a2.xlsx"]);
}

// Returns empty, no error thrown
const results = await Promise.all([fun1(), fun2()]);
```

## Testing

- [x] Identical SQL strings in `Promise.all` now return correct results
- [x] Different SQL strings still work as before  
- [x] Sequential (non-concurrent) queries unaffected
- [x] Queries without cache (`alasql.options.cache = false`) unaffected

### DIFF
--- a/src/17alasql.js
+++ b/src/17alasql.js
@@ -261,10 +261,11 @@ function cleanupCache(statement) {
 	if (!alasql.options.cache) {
 		return;
 	}
-	// cleanup the table data to prevent storing this information in the SQL cache
-	if (statement && statement.query && statement.query.data) {
-		statement.query.data = [];
-	}
+	// NOTE: Do NOT mutate statement.query.data on the cached statement object.
+	// The compiled statement is shared across concurrent executions via sqlCache.
+	// Mutating it here clears data that another concurrent Promise.all call
+	// may still be reading, causing empty results with no error thrown.
+	// See: https://github.com/AlaSQL/alasql/issues/1953
 }
 
 /**
@@ -284,7 +285,9 @@ alasql.dexec = function (databaseid, sql, params, cb, scope) {
 		let statement = db.sqlCache[hh];
 		// If database structure was not changed since last time return cache
 		if (statement && db.dbversion === statement.dbversion) {
-			var res = statement(params, cb);
+			// Clone params to prevent concurrent executions from sharing
+			// the same params reference when called via Promise.all
+			var res = statement(Array.isArray(params) ? params.slice() : Object.assign({}, params), cb);
 			cleanupCache(statement);
 			return res;
 		}


mathiasrw:
Ohhh, very nice. 

Please make a file named `/test/test-1953.js` where we demonstrate the issue and will catch if we ever accidently introduce the error again.

src/17alasql.js
```diff
@@ -261,10 +261,11 @@ function cleanupCache(statement) {
...
 	}
-	// cleanup the table data to prevent storing this information in the SQL cache
-	if (statement && statement.query && statement.query.data) {
-		statement.query.data = [];
-	}
+	// NOTE: Do NOT mutate statement.query.data on the cached statement object.
+	// The compiled statement is shared across concurrent executions via sqlCache.
+	// Mutating it here clears data that another concurrent Promise.all call
+	// may still be reading, causing empty results with no error thrown.
+	// See: https://github.com/AlaSQL/alasql/issues/1953
```
mathiasrw, line 268: This is informative for this PR but not useful after this has been merged. Please remove.

mathiasrw (changes requested):
Please provide a test file named test/test1953.js that fails with the original code and that does not failed with the modified code.

mathiasrw (changes requested):
We need a test file that fails before the merge and passes after the merge.

mathiasrw:

PR #2479, opened 2026-04-19
Fix MIN() to handle date comparison correctly

Aston8:


gabrielmongefranco:
Hello, I just noticed this pull request. Please see #2481 and #2480.

mathiasrw:
Thank you for this. 

How can we be sure we did not break something?

mathiasrw:
Any chance you can add a test file?

gabrielmongefranco:
Hi, let me dig through the git history in my repo to understand the issue better. I remember I had to do some work arounds but I can't remember why. :)

PR #2480, opened 2026-04-19
Refactor MAX and MIN functions to handle null values

Aston8:


gabrielmongefranco:
Hi, I just noticed this pull request. Please also see #2481.

Aston8:
Thanks! I’ll review #2481 and update this PR accordingly.

mathiasrw:
Thank you for this. 

How can we be sure we did not break something?

mathiasrw:
Any chance you can add a test file?

gabrielmongefranco:
HI, sorry for the late reply! I can try... :)

PR #2521, opened 2026-08-04
Fix SUBSTRING with a non-positive start position

spokodev:
## Problem

`SUBSTRING` / `SUBSTR` / `MID` compile `SUBSTRING(str, pos, len)` to `str.substr(pos - 1, len)`. When `pos` is `0` or negative, `pos - 1` becomes a negative index, and JavaScript's `String.prototype.substr()` counts a negative start from the **end** of the string. The result is neither correct nor consistent with any SQL dialect:

```js
alasql("SELECT VALUE SUBSTRING('abcdef', 0, 3)")   // 'f'   (MySQL: '')
alasql("SELECT VALUE SUBSTRING('abcdef', -1, 3)")  // 'ef'  (MySQL: 'f')
alasql("SELECT VALUE SUBSTRING('abcdef', -2, 3)")  // 'def' (MySQL: 'ef')
```

So a computed start position that lands on `0` or goes negative silently returns wrong characters.

## Fix

Follow MySQL semantics (verified against MySQL 8), which alasql already matches for positive positions:

- start `= 0` yields an empty string;
- a negative start counts from the end (`SUBSTRING('abcdef', -2)` -> `'ef'`), clamped to an empty string when it reaches past the beginning (`SUBSTRING('abcdef', -7)` -> `''`);
- positive start positions are unchanged.

| call | before | after (= MySQL 8) |
| --- | --- | --- |
| `SUBSTRING('abcdef', 0, 3)` | `'f'` | `''` |
| `SUBSTRING('abcdef', -1, 3)` | `'ef'` | `'f'` |
| `SUBSTRING('abcdef', -2, 3)` | `'def'` | `'ef'` |
| `SUBSTRING('abcdef', -2)` | `'def'` | `'ef'` |
| `SUBSTRING('abcdef', -7, 3)` | `'bcd'` | `''` |
| `SUBSTRING('abcdef', 2, 3)` | `'bcd'` | `'bcd'` |

## Test

Added `test/test-substring-negative-start.js`. `yarn test` passes (2193 tests).

### DIFF
--- a/src/55functions.js
+++ b/src/55functions.js
@@ -186,24 +186,8 @@ stdlib.SUBSTRING =
 	stdlib.SUBSTR =
 	stdlib.MID =
 		function (a, b, c) {
-			if (arguments.length == 2)
-				return und(
-					a,
-					'(__alasql_tmp=(' +
-						b +
-						'),__alasql_tmp==0?"":(__alasql_tmp<0?(y.length+__alasql_tmp<0?"":y.substr(y.length+__alasql_tmp)):y.substr(__alasql_tmp-1)))'
-				);
-			else if (arguments.length == 3)
-				return und(
-					a,
-					'(__alasql_tmp=(' +
-						b +
-						'),__alasql_tmp==0?"":(__alasql_tmp<0?(y.length+__alasql_tmp<0?"":y.substr(y.length+__alasql_tmp,' +
-						c +
-						')):y.substr(__alasql_tmp-1,' +
-						c +
-						')))'
-				);
+			if (arguments.length == 2) return und(a, 'alasql.utils.substr(y,' + b + ')');
+			else if (arguments.length == 3) return und(a, 'alasql.utils.substr(y,' + b + ',' + c + ')');
 		};
 
 stdfn.REGEXP_LIKE = function (a, b, c) {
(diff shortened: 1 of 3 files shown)


src/55functions.js
```diff
@@ -186,8 +186,24 @@ stdlib.SUBSTRING =
...
 	stdlib.MID =
 		function (a, b, c) {
-			if (arguments.length == 2) return und(a, 'y.substr(' + b + '-1)');
-			else if (arguments.length == 3) return und(a, 'y.substr(' + b + '-1,' + c + ')');
+			if (arguments.length == 2)
+				return und(
+					a,
+					'(__alasql_tmp=(' +
+						b +
+						'),__alasql_tmp==0?"":(__alasql_tmp<0?(y.length+__alasql_tmp<0?"":y.substr(y.length+__alasql_tmp)):y.substr(__alasql_tmp-1)))'
```
mathiasrw: I feel like we are introducing complexity by writing it like this. Are you able to make it a bit easier to understand what is happening?

mathiasrw (changes requested):
I really love that you fix something like this. 

This will be a breaking change and will be included in the next major version update.

